PR #158, opened 2016-09-16
SearchTerm integration and miscelaneous fixes

cjfreeze:
This PR contains the first implementation of the search_terms feature, as well as minor fixes and optimizations.


wintermeyer:
`Dev` is not a good name for a pull request. It should have a little bit more information so that I get an idea what it is about.

bildschirmfoto 2016-09-17 um 08 22 40

wintermeyer:
The results of a search should be sorted by not only `similarity` but also by `last_name` + `first_name`.

wintermeyer (changes requested):
`score` might be a better field name name `similarity`. My bad.

wintermeyer (changes requested):
The results of a search should be sorted by not only `similarity` but also by `last_name` + `first_name`.

lib/vutuv/registration.ex
```diff
@@ -2,30 +2,49 @@ defmodule Vutuv.Registration do
 	import Ecto.Query
 	alias Vutuv.User
 	alias Vutuv.Slug
-	alias Vutuv.Repo
+  alias Vutuv.Repo
```
wintermeyer, line 5: Wrong formatting. In case you are mixing tabs and spaces: Don't!
**Only use spaces!**

lib/vutuv/registration.ex
```diff
@@ -2,30 +2,49 @@ defmodule Vutuv.Registration do
...
+    end)
+   	Repo.insert(changeset)
+	end
+
+	def generate_search_terms(%{"first_name" => first_name,
+                              "last_name" => last_name} = user) do
+		terms = 
+      [first_name,
+      last_name,
+      user["emails"]["0"]["value"],
```
wintermeyer, line 35: In case a user has more than one e-mail address the other e-mail addresses should be added too.

web/controllers/search_query_controller.ex
```diff
@@ -51,4 +50,12 @@ defmodule Vutuv.SearchQueryController do
 
   def query_changeset(changeset, _, _), do: changeset
 
+  def search(value) do
+    for(term<- Repo.all(from t in Vutuv.SearchTerm, where: ^value == t.value)) do
+      %{similarity: term.similarity, user_id: term.user_id}
+    end
+    |>Enum.sort(&(&1.similarity> &2.similarity)) #sorts by similarity
```
wintermeyer: Should sort by `last_name` and `first_name` in addition to the `similarity`.

wintermeyer (changes requested):
I like the comments. I only have a couple of small fixes.

wintermeyer (changes requested):
search_terms are not downcased.

```
mysql> select value from search_terms;
+-------------------------------+
| value                         |
+-------------------------------+
| Stefan                        |
| Wintermeyer                   |
| stefan.wintermeyer@amooma.de  |
| Stefan Wintermeyer            |
| Wintermeyer Stefan            |
| Stefan, Wintermeyer           |
```

They should be downcased. And the search should downcase the search requests.

wintermeyer:
José Valim should have a `search_terms` with and without the "é".
- José Valim
- Jose Valim

lib/vutuv/registration.ex
cjfreeze: I think this one can be blamed on my text editor. I recently upgraded to sublime text 3, and it seems to have trouble remembering my indentation settings for multiple files. I've been looking into it.

web/controllers/search_query_controller.ex
cjfreeze: So `users` with equal `scores` will be alphabetized by `last_name`?

wintermeyer: Yes, they should be alphabetized by `last_name` plus `first_name`.

web/router.ex
```diff
@@ -50,6 +50,7 @@ defmodule Vutuv.Router do
       resources "/work_experiences", WorkExperienceController
       resources "/addresses", AddressController
       resources "/oauth_providers", OAuthProviderController
+      resources "/search_terms", SearchTermController
```
wintermeyer: Shouldn't this be followed by a `only: [:new, :create, :show, :delete, :index]`?

wintermeyer (changes requested):
The `search_queries` resource doesn't feel RESTful yet. After a `search_queries#new` it should lead to an `search_queries#show` (with the `id` or the `slug`). Now it leads to `search_queries#index` which doesn't make sense in the RESTful world. One can argue that a search results in a score of 100 and a single entry than we could do a redirect to that user. But that's a second step. Right now I'd like `search_queries` behave like a proper RESTful resource.

wintermeyer:
`search_queries#index` should only display the `search_queries` of the current user. If no user is logged in it should redirect to `search_queries#new`.

PR #481, opened 2016-12-03
Include docker support for development environments

BrummbQ:
Hi,
I was missing a quick start setup for vutuv, so I included docker support. Now you can build and run vutuv with just one command (tested on linux). This could lower the barrier for new developers, also you keep your system clean.

Hope you like it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+_build
+node_modules
+deps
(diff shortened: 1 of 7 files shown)


wintermeyer:
WOW! Thank you! I just asked our Twitter feed for feedback. We're busy fixing bugs so we have to rely on the community in this one.

PR #484, opened 2016-12-05
Validate url

BrummbQ:
Hello again,

To prevent any gibberish url input, I tried to improve the validation. The model first parses the url and then checks the hostname, so this should be quite safe. To test the new check I implemented a test for the url controller and on the way also upgraded and used ex machina, which was working surprisingly pleasant.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -57,7 +57,7 @@ defmodule Vutuv.Mixfile do
      {:cowboy, "~> 1.0"},
      {:arc, "~> 0.5.3"},
      {:arc_ecto, "~> 0.4.2"},
-     {:ex_machina, "~> 0.6.1"},
+     {:ex_machina, "~> 1.0"},
      {:phoenix_html_simplified_helpers, "~> 0.6.0"},
      {:mariaex, ">= 0.0.0"},
      {:word_smith, "~> 0.1.0"},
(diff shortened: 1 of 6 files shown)


cjfreeze:
Thanks for taking the time to try out the app!

Firstly, I have a question, and I admit beforehand that I only glanced over the changes.
My question is: what was the purpose of the changes of in the templates?

Secondly, I see that you generated tests from ex machina for this. I've been looking into that for when we get to the milestone that addresses tests. For now though, we'd like to leave tests untouched... But feel free to separate them into another pull request.

wintermeyer:
@cjfreeze The templates haven't changed. AFAIK they are just nicer intended.

wintermeyer:
I haven't tested it yet but I like it so far. I'll give it a try tomorrow.

wintermeyer:
But I would have liked it to be two PRs.

cjfreeze:
Ah just cleaned the templates then. Yeah that does need to be done in several areas.

wintermeyer:
I checked the Pull-Request and I appreciate the effort. Thank you!

Please resolve the following issues:
- `can't resolve url` needs to be translated into the German `die URL kann vom Nameserver nicht aufgelöst werden` (please contact us in case you don't know how to do I18n work)
- The correct `mix.lock` is missing. We can not deploy without the correct version.
- Please split up the PR into two. One for the ex_machina upgrade and one for the url validation.

wintermeyer:
https://github.com/vutuv/vutuv/blob/master/faster_pr_reviews.md

PR #508, opened 2016-12-19
fix typo

mfa:


wintermeyer:
Thanks for the pull request! A couple of more files need to be handled to fix this issue. To save you some time I fixed it in this commit: https://github.com/vutuv/vutuv/commit/a9f5d43662f179de31f5529cc94cac8d43d24163

PR #509, opened 2016-12-23
Add a locale database model

cjfreeze:
A `Locale` database model allows a few different things: Firstly it makes any locale database field cheaper, as all it needs to store is an int rather than a string. Secondly it helps with localization because of the way `exonyms` can vary depending on the user's locale. Thirdly it makes forms easier, because helper functions are provided to output lists to be passed to Phoenix's `select` form helper.

This PR includes a migration to generate the contents of the `locale` table and the `endonyms` for each locale. It does not include `exonyms`. It includes an admin page to add exonyms as necessary.

## Example dataset

`de` is the locale code for `Deutsch`. The endonym of `de` is `Deutsch`. There are many `exonyms` for `Deutsch`. For example, the `English` (locale code `en`) exonym for `de` is `German`. For `French` locale code `fr`) the eponym for ` de` is `Allemand`.

See https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

## Screenshot

image

The above screenshot shows an additional example of the `Japanese` exonym for `Deutsch` being added via the embedded form.

image

The English exonym for Deutsch is shown in the above screenshot.


wintermeyer (changes requested):
- The breadcrumbs should represent, that we are in the admin area.
- The PR is not clean. Many incidence of '#' for future to be code. Those lines shouldn't be in the code at all. Who knows what happens in the future. The PR has to be good itself.
- What is the solution for languages we don't have in the database? Is there an "other" language?

wintermeyer (approved):

PR #515, opened 2017-01-03
Add a Tag database model

cjfreeze:
`Tags` are a system that allow you to define yourself however you like. Anything can be a tag, an interest, a skill, a hobby, etc. This PR adds a tag system to vutuv.

Tags themselves only have one field: `slug`.

image

The slug is how they appear in the browser, and has to be unique.

image

Tags have two one to many relationships:

`tag_localizations`:

image

And `tag_synonyms`:

image

Localizations allow tags to be presented in many different languages, such as the tag `SQL` for English:

image

And German:

image

Localizations have a `name`, `description`, and `locale`.

Localization's are unique between their locale and parent tag, so only one localization per tag per locale is allowed.
Localizations also have a one to many relationship: `tag_urls`, so tags can have several urls.

image

Urls have a `value` (for the link), a `name`, and a `description`

Because URLS belong to a localization, they can be different depending on the user's locale.

image

Synonyms are being added in this pull request for convenience, but most of their functionality will be added in the future `user_tags` pull request, which will allow users to add tags to themselves. For now they can be added, and are unique, but will not disallow you from creating a tag with the same name. This is intended behavior, as the admin panel is the only way to currently create tags, and tags created by the admin panel should not be subject to the synonym auto-change. Only user_skills will interact with synonyms, and they are yet to be added.


wintermeyer (changes requested):
This PR contains stuff which is already in the master. e.g. `priv/repo/migrations/20161221000716_populate_locales.exs`
Please clean it up.

web/models/tag.ex
```diff
@@ -0,0 +1,89 @@
...
+  use Vutuv.Web, :model
+  @derive {Phoenix.Param, key: :slug}
+  
+
+  schema "tags" do
+    field :slug, :string
+
+    has_many :tag_localizations, Vutuv.TagLocalization, on_delete: :delete_all
+    has_many :tag_synonyms, Vutuv.TagSynonym, on_delete: :delete_all
+    #has_many :user_tags, Vutuv.UserTag, on_delete: :delete_all
```
wintermeyer: Never leave those comments in the code. Keep the PR clean.

cjfreeze: Ah I forgot one, my bad

wintermeyer (changes requested):
I like the PR. The only thing which I'd like to change is the default naming convention for slugs. Instead of a "." a "_" makes more sense. e.g. "ruby_on_rails" instead of "ruby.on.rails".

PR #520, opened 2017-01-11
Add Pin login system for use by Facebook Free Basics users.

cjfreeze:
This PR adds additional functionality to magic links in the form of temporary login `Pins`. 

This system is inconsequential to anyone who is not sending the header `X-IORG-FBS = true`
If you have that header (usually when you are a Facebook Free Basic user) the user login and account creation change to use pins instead of magic links. 

In the screenshot below, I am using a utility to have the required header, and created an account. This is the pin entry page that is displayed. 

image

The screenshot below is the page that is displayed when logging in, or if there is a pin session cookie stored and the user attempts to visit the vutuv index page.

image

The email is also changed accordingly, and looks like this:

```
Hi Test,

Thank you for joining vutuv. Enter the following pin on vutuv to login
849570
This link expires in 30 minutes. You can always get a new one by
returning to vutuv and logging in.

If you like our service, feel free to share us with your friends!

Regards
The vutuv team
```

Or for just logging in, this:

```
Hi Test,

Enter the following pin on vutuv to login
667677
Thanks for using vutuv! If you like our service, feel free to share us
with your friends!

Regards
The vutuv team

PS: The pin expires in 30 minutes. You can always get a new one by
returning to vutuv and logging in.
```

Some things to note

This system might require further testing in a production environment, as there may be some technicalities with how the header is actually provided.

There is also some existing issues with the cookie being used to store the client-side portion of the authentication process, where it is not deleted when it should be. It does expire in 30 minutes, and server-side fixes have been made to prevent this issue from occurring too often.

Some possible changes that could be made before the system is merged includes using the session instead of an independent cookie, and using server-side checks to make sure the pin has not expired when reading the session and deciding to display the pin form.

Finally, currently the encrypted cookie is storing the user's email for use in pin security, but a possible better alternative would be the magic link hash. 

These things all need to be discussed before the PR can be merged.


lib/vutuv/magic_link_helpers.ex
```diff
@@ -36,6 +40,12 @@ defmodule Vutuv.MagicLinkHelpers do
     |> String.downcase
   end
 
+  defp gen_pin do
```
wintermeyer, line 43: A PIN should not start with a 0.

PR #528, opened 2017-01-24
Migrate skills to tags

cjfreeze:
With the tag system in full effect, all that is left is to migrate all existing skills to tags. This PR does this for both the back end and the front end.


wintermeyer:
This click:
screenshot at jan 24 20-34-04

Leads to this error:
bildschirmfoto 2017-01-24 um 20 32 41

wintermeyer:
It's impossible for a new user to create new tags with the easy entry form. 

bildschirmfoto 2017-01-26 um 14 18 41

PR #575, opened 2019-03-02
schema user, email_address, authentication, session

hendri-tobing:
Schema for user, email_address, authentication, session. Using phauth, arc, and other dependencies.


lib/vutuv_web/uploaders/avatar.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  def sym_link_for_avatar_exists?(user) do
+    timestamp = user.updated_at
+    |> NaiveDateTime.to_string
+    |> String.replace(~r/[^0-9]/,"")
+
+    nginx_path = user.active_slug
+    |> String.replace(".","/")
+
+    sym_link_path = "/var/www/vutuv/avatars/#{nginx_path}/#{timestamp}"
```
wintermeyer, line 46: Please define a variable for `/var/www/vutuv/` which can changed in `config/[prod|dev].exs`

Just one place to update this kind of information when needed.

lib/vutuv/Schema/user.ex
```diff
@@ -0,0 +1,77 @@
...
+      field :validated?, :boolean, default: false
+  
+      has_many :email_addresses,  Vutuv.EmailAddress, on_delete: :delete_all
+  
+      timestamps()
+    end
+  
+    def changeset(user, attrs) do
+      user
+      |> cast(attrs, [:first_name, :last_name, :middlename, :nickname, :honorific_prefix, :honorific_suffix, :gender, :birthday_day, :birthday_month, :birthday_year, :locale, :verified, :avatar, :active_slug, :headline, :noindex?, :validated?])
```
wintermeyer: `:active_slug, :noindex?, :validated?` shouldn't be in this cast or should they? We don't want a user to inject those values in the WebGUI.

lib/vutuv/Schema/user.ex
```diff
@@ -0,0 +1,77 @@
...
+    defp validate_first_name_or_last_name(changeset, %{}) do
+      first_name = get_field(changeset, :first_name)
+      last_name = get_field(changeset, :last_name)
+  
+      if first_name || last_name do
+        # No error if any of those 2 are present.
+        #
+        changeset
+      else
+        # All the 3 fields are nil.
```
wintermeyer, line 67: Shouldn't this comment say "All 2 fields are nil"?

lib/vutuv/Schema/user.ex
```diff
@@ -0,0 +1,77 @@
...
+        #
+        message = "First name or last name must be present"
```
wintermeyer, line 69: We have to have a I18n solution for these messages. We can develop the application with the english texts but have to have an easy option to add translations (German being the first).

wintermeyer (changes requested):
Just some small things. I added comments.

lib/vutuv/Schema/user.ex
riverrun: The translations will be handled in the view layer. We just need to make that the message is correctly added to the changeset, as it is here, and that we update the pot (gettext) files with each message.

hendri-tobing:
Close this PR #575 override by #580

Issue #573, opened 2019-03-01
EmailAddress Scaffold


wintermeyer:
A `user` has_many `email_addresses`. 

# Attributes

- user_id, :integer
- value, :string
- description, :string
- is_public, :boolean, default: true
- position, :integer

## Indexes

- value (unique)
- user_id

## Validations

- @required_fields ~w(value is_public user_id)
- The associated user has to be valid.
- value has to be a valid email address. But don't go overboard with this. A 99% validation is fine for now.

## Misc

- Within a `user` all `email_addresses` are always ordered by the `position`.
- `value` is auto downcased during create and update.

# Routing

They are nested under the `users` resource:
`/users/:slug/email_addresses`

# Form

- value, :string
- description, :string # Pulldown select with 'Work', 'Home' and a free field where people can enter they own text.
- is_public, :boolean, default: true

riverrun:
Will every email address need to be verified? Do we want to add a `verified` attribute to the table?

hendri-tobing:
> Will every email address need to be verified? Do we want to add a `verified` attribute to the table?

I think so, to minimize false email addresses and security reason.

wintermeyer:
> > Will every email address need to be verified? Do we want to add a `verified` attribute to the table?
> I think so, to minimize false email addresses and security reason.

Yes, we do have to verify every new email address by sending one email to it and having the user clicking on the verification link. We already do it in v1.1. Otherwise I could add your email address to my account and you couldn't use it any more.

Issue #570, opened 2019-02-28
User Scaffold


wintermeyer:
# Introduction

The `user` model is the core of vutuv. We have to import all the data of v1.1 to the new model and should try to clean it up as good as possible during the process of creating a v2.0 version. The import process will be dealt with in an extra issue.

## Authentication

Obviously the `user` model goes hand in hand with authentication. I think [phauxth](https://github.com/riverrun/phauxth) is a good library for Phoenix we can use. But I'm open for other solutions. We are not going to have a magic-link-only login procedure but a password first and if somebody doesn't want it magic-link second solution (for which we can use the phauxth password reset mechanism). So old users can still log in without a password.

# Attributes

- :first_name, :string
- :last_name, :string
- :middlename, :string
- :nickname, :string
- :honorific_prefix, :string
- :honorific_suffix, :string
- :gender, :string # male, female, other (not sure if other or diverse is a better fit)
- :birthday_day, :integer
- :birthday_month, :integer
- :birthday_year, :integer
- :locale, :string # de|en (currently)
- :verified, :boolean, default: false
- :avatar, Vutuv.Avatar.Type
- :active_slug, :string
- :headline, :string
- :noindex?, :boolean, default: false
- :validated?, :boolean, default: false
- :send_birthday_reminder, :boolean, default: true
- :password_digest/hash # other name also fine

## Validations

- There has to be a :first_name or a :last_name.
- There has to be a :gender # please check v1.1 for the values
- There has to be a :local [de|en]
- There has to be a :active_slug which has to be generated during the create.
- The birthday is a bit tricky. We want people to enter either their full birthday (including the year) or just the day and month (without the year) in case they don't want anybody to know how old they are. Validation has to cover this. But a birthday doesn't have to be entered.
- validate_length(:first_name, max: 80)
- validate_length(:last_name, max: 80)
- validate_length(:middlename, max: 80)
- validate_length(:nickname, max: 80)
- validate_length(:honorific_prefix, max: 80)
- validate_length(:honorific_suffix, max: 80)
- validate_length(:headline, max: 255)

## Questions/Problems

- In the old model we did not have an `email_address` field. It was separated in an extra model to have the possibility of several email addresses. I think this is still a valid approach but I'm not 100% sure any more. Feel free to include it here too if phauxth needs it. If we put `email_address` in this model we have to add a unique restrain in the validations. And email addresses should always be saved downcased.
[...]

PR #580, opened 2019-03-07
some updates

hendri-tobing:
Sorry, I create a new PR from a fork in my repository for some reasons.

This is, some update of issue #573  and #570. Sorry for the late update, because I lost my work on local for couple days.


riverrun:
Does this PR override #575? If it does, could you close #575? So that it does not get merged by mistake.

hendri-tobing:
> Does this PR override #575? If it does, could you close #575? So that it does not get merged by mistake.

Done, thank you.

config/test.exs
```diff
@@ -16,3 +16,13 @@ config :vutuv, Vutuv.Repo,
   database: "vutuv2_test",
   hostname: "localhost",
   pool: Ecto.Adapters.SQL.Sandbox
+
+
+# Password hashing test config
+config :argon2_elixir, t_cost: 1, m_cost: 8
+#config :bcrypt_elixir, log_rounds: 4
```
riverrun: Let's delete the commented lines - we don't need them.

lib/vutuv/accounts/roles.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Vutuv.Accounts.Role do
```
riverrun: Let's change the name of the file to `role.ex` - not `roles.ex`

lib/vutuv/biographies/profile.ex
```diff
@@ -0,0 +1,75 @@
...
+    |> validate_length(:first_name, max: 80)
+    |> validate_length(:last_name, max: 80)
+    |> validate_length(:middlename, max: 80)
+    |> validate_length(:nickname, max: 80)
+    |> validate_length(:honorific_prefix, max: 80)
+    |> validate_length(:honorific_suffix, max: 80)
+    |> validate_length(:headline, max: 255)
+  end
+
+  def gender_gettext("male") do
```
riverrun: Why are we calling gettext from here?

wintermeyer:
@hendri-tobing I added @riverrun and I as a reviewer but actually don't know if you are finished with this PR. Is it ready for review?

lib/vutuv/accounts.ex
```diff
@@ -0,0 +1,290 @@
...
+  ## Examples
+
+      iex> get_email_address!(123)
+      %EmailAddress{}
+
+      iex> get_email_address!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_email_address!(id), do: Repo.get!(EmailAddress, id)
```
riverrun: The phoenix generators create bang functions for the 'get_item' function. I think the regular versions (non-bang) are more useful, but what do you think?

lib/vutuv/accounts/email_address.ex
```diff
@@ -0,0 +1,25 @@
...
+
+    timestamps()
+  end
+
+  def changeset(email_address, attrs) do
+    email_address
+    |> cast(attrs, [:value, :user_id])
+    |> validate_required([:value])
+    |> validate_format(:value, ~r/^[A-Za-z0-9\._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,6}$/)
+    |> validate_length(:value, max: 80)
```
riverrun: Do we want to limit emails to 80 chars? My understanding is that emails can be anything up to 255 chars long.

lib/vutuv/accounts/user.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  schema "users" do
+    field :email, :string
+    field :password, :string, virtual: true
+    field :password_hash, :string
+    field :confirmed_at, :utc_datetime
+    field :reset_sent_at, :utc_datetime
+
+    has_many :sessions, Session, on_delete: :delete_all
+    has_one :profiles, Profile, foreign_key: :user_id, on_delete: :delete_all
```
riverrun: With `has_one`, let's use `:profile`, not `:profiles`, and I don't think we need the `foreign_key` entry - it will be user_id by default.

lib/vutuv/accounts/user.ex
```diff
@@ -0,0 +1,86 @@
...
+    many_to_many :roles, Role, on_delete: :delete_all, join_through: "user_roles"
+
+    timestamps()
+  end
+
+  def changeset(%__MODULE__{} = user, attrs) do
+    user
+    |> cast(attrs, [:email])
+    |> validate_required([:email])
+    |> unique_email
```
riverrun: We won't need the `unique_email` calls in this module if we are checking the emails in the email_addresses table, and ideally we will only be checking for uniqueness in one place.

We could replace this line with `|> cast_assoc(:email_addresses)`

lib/vutuv_web/controllers/user_controller.ex
```diff
@@ -0,0 +1,92 @@
...
+  def create(conn, %{"user" => %{"email" => email} = user_params}) do
+    key = Token.sign(%{"email" => email})
+    case Accounts.create_user(user_params) do
+      {:ok, user} ->
+        Log.info(%Log{user: user.id, message: "user created"})
+
+        Email.confirm_request(email, key)
+
+        #create profile
+        profile = Ecto.build_assoc(user, :profiles, %{gender: user_params["gender"], first_name: user_params["first_name"], last_name: user_params["last_name"]})
```
riverrun: All this logic (including nearly all Ecto and Repo calls) should be in the relevant context module.

lib/vutuv_web/controllers/user_controller.ex
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+  def show(%Plug.Conn{assigns: %{current_user: user}} = conn, %{"id" => id}) do
+    user = if id == to_string(user.id), do: user, else: Accounts.get_user(id)
+    
+    profile = if id == to_string(user.id) do 
+      profile = Biographies.get_profile!(id) 
+    else 
+      Repo.get!(Profile, id)
```
riverrun: Again, we should not need to call Repo directly from the controllers

lib/vutuv_web/controllers/user_controller.ex
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    post = if id == to_string(user.id) do
```
riverrun: The id check is being done three times in this function. We should only need to do it once.

lib/vutuv_web/router.ex.bak
```diff
@@ -0,0 +1,26 @@
+defmodule VutuvWeb.Router do
```
riverrun: We can remove all of the `.bak` files - they are just backups created by the Phauxth installer.

lib/vutuv_web/views/user_view.ex
```diff
@@ -0,0 +1,17 @@
+defmodule VutuvWeb.UserView do
+  use VutuvWeb, :view
+  alias VutuvWeb.UserView
```
riverrun, line 3: We don't need these Json routes, do we?

lib/vutuv/accounts/user.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  import Ecto.Changeset
+
+  alias Vutuv.Sessions.Session 
+  alias Vutuv.Accounts.EmailAddress
+  alias Vutuv.Accounts.Role
+  alias Vutuv.Biographies.Profile
+
+  schema "users" do
+    field :email, :string
```
riverrun: As email_addresses are a separate table, I don't think we need this entry.

priv/repo/migrations/20190307081841_create_users.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Vutuv.Repo.Migrations.CreateUsers do
+  use Ecto.Migration
+
+  def change do
+    create table(:users) do
+      add :email, :string
```
riverrun: As email_addresses are a separate table, I don't think we need this entry - see comment for schema above.

riverrun (changes requested):
I have just left a few comments - together with a couple of questions.

lib/vutuv/accounts.ex
hendri-tobing: Agree. The bang version function will cause crash the app if it does not work

lib/vutuv/accounts/email_address.ex
hendri-tobing: Ok. I will change the length to 255.

lib/vutuv_web/views/user_view.ex
hendri-tobing: Yes, we don’t need it now.

hendri-tobing:
Thank you for the review. I will fix them as soon as possible.

wintermeyer (commented):
I'm not a big fan of the `some updates` title of this PR. ;-)

The emails which are send include "http://www.example.com" in development. Is there a possibility to make that http://localhost:4000 to make it a bit easier for new developers. It is not a show stopper but would be nice.

And the README needs an update of how the URL for production can be configured.

hendri-tobing:
> I'm not a big fan of the `some updates` title of this PR. ;-)
> 
> The emails which are send include "http://www.example.com" in development. Is there a possibility to make that http://localhost:4000 to make it a bit easier for new developers. It is not a show stopper but would be nice.
> 
> And the README needs an update of how the URL for production can be configured.

I will change the title and the email config for development

riverrun (approved):

riverrun:
@wintermeyer shall we merge this now? I'm happy with the changes. Once it is merged, I can rebase #584 and get that merged

wintermeyer:
@riverrun Go ahead with the merge.

Issue #571, opened 2019-02-28
Avatar


wintermeyer:
Add an avatar to the `user` model. In v1.1 we used https://github.com/stavro/arc but I'm open for other solutions if they make more sense. The files should be stored locally on the file system of the server.

Please have a look at the v1.1 version how we populated the avatars in the sign_up process by looking up the email addresses at https://www.gravatar.com. Feel free to copy and paste that code or to reinvent the wheel.

hendri-tobing:
I will create a PR for this issue later today.

riverrun:
The avatar has been added to the user model. The remaining work is to add a task to fetch the image from gravatar - this will be handled in #611.

PR #602, opened 2019-04-01
update profile with arc and arc_ecto, profile templates and route

hendri-tobing:
Update the following: 
* Profile's avatar field with arc and arc_ecto (initial work on #571 ), 
* Profile templates, 
* Profile route.

When testing with arc_ecto, there are some problems:
* It does not work with Dialyzer, [reference: stackoverflow](https://stackoverflow.com/questions/49511255/typespecs-for-arc-ecto-changeset) 
* It insert DateTime on updated_at field with microseconds. This is causing many tests fail, [the issue: here](https://github.com/stavro/arc_ecto/issues/96)

For a temporary solution, I forked arc_ecto and update the necessary things to pass the above issues. I add this on the `mix.exs`

`{:arc_ecto, git: "https://github.com/hendri-tobing/arc_ecto.git"}`


wintermeyer (dismissed):
I'm not the right person to judge this PR but I approve it to keep things going. Obviously the self hosted arc_ecto is suboptimal. 

@riverrun Please share your thoughts when you are ready.

lib/vutuv/biographies/profile.ex
```diff
@@ -27,6 +28,7 @@ defmodule Vutuv.Biographies.Profile do
           updated_at: DateTime.t()
         }
 
+  @timestamps_opts [type: :naive_datetime, usec: false]
```
riverrun: Why do we need this?

riverrun (commented):
Just one question.

Also, the travis build is failing with this error:

warning: function validate_file/1 is unused
  lib/vutuv_web/uploaders/avatar.ex:33
Compilation failed due to warnings while using the --warnings-as-errors option
The command "mix compile --warnings-as-errors" exited with 1.

lib/vutuv/biographies/profile.ex
hendri-tobing: No we don't need it. Thank you for the review.

riverrun:
Looks good.

riverrun (approved):

PR #609, opened 2019-04-13
Update gravatar as a default avatar

hendri-tobing:
When registering as a new user, it is automatically getting a default avatar from https://www.gravatar.com as described in #571 . If the new user does not have it on gravatar, the default avatar will be the vutuv logo.


lib/vutuv/accounts.ex
```diff
@@ -282,6 +286,48 @@ defmodule Vutuv.Accounts do
...
+      |> :erlang.md5()
+      |> Base.encode16(case: :lower)
+
+    Application.ensure_all_started(:inets)
+    Application.ensure_all_started(:ssl)
+
+    img = 'https://www.gravatar.com/avatar/#{hash}?s=150&d=404'
+
+    if {:ok, {status, header, body}} = :httpc.request(:get, {img, []}, [], []) do
+      case status do
```
riverrun, line 309: This case can be simplified - for example:

```
case status do
  {_, 200, 'OK'} -> create_upload
  _ -> nil
end
```

lib/vutuv/accounts.ex
```diff
@@ -282,6 +286,48 @@ defmodule Vutuv.Accounts do
...
+          nil
+
+        {_, 200, 'OK'} ->
+          content_type = find_content_type(header) |> to_string
+          filextension = String.replace(content_type, "image/", "")
+          filename = "original.#{filextension}"
+          path = "#{Application.get_env(:vutuv, :storage_dir)}" <> "#{userid}/"
+          File.write(path <> filename, body)
+
+          _upload = %{
```
riverrun: We don't need `_upload = ` here - just return the map.

lib/vutuv/accounts.ex
```diff
@@ -80,7 +82,9 @@ defmodule Vutuv.Accounts do
 
     case Repo.insert(user_with_email) do
       {:ok, new_user} ->
-        user_with_profile = Ecto.build_assoc(new_user, :profile, profile_attrs)
+        img = get_gravatar(attrs["email"], new_user.id)
```
riverrun, line 85: What happend when this function returns nil?

riverrun (changes requested):
Looks good - just a couple of small comments and a question.

lib/vutuv/accounts.ex
hendri-tobing: If the get_gravatar return nil, the avatar will be the default avatar of Vutuv (located at assets/static/images/avatars/)

hendri-tobing:
I have updated as advised. Please let me know if anything is missing. Thank you.

riverrun (approved):

Issue #613, opened 2019-04-29
[QUESTION] Email address controllers


riverrun:
This issue is just about clarifying how we want the email_address controllers to behave. I will outline my thoughts here, and then you can comment on them below.

## Routes

* index - lists all the email_addresses for a specific user (the current user)
* show - lists all the details for a specific email_address
* new / create - form to create a new email_address
    * this will need to include a function to request email confirmation
* edit / update - form to update an email_address
    * the user should not be able to update the `value` this way (updating the `value` needs to be done by deleting the email_address and creating a new one)

## Authorization

All the email_address controller routes should be protected. The current user should only be able to access his / her own email_address resources. None of the routes should be open to the public.

wintermeyer:
I agree with everything. Maybe we should add a 404 with a specific error message for the edit/update. So that a user understands why he/she can not edit an email address.

riverrun:
One more question - shall we nest the routes, so that the route would be, for example, `/users/1/email_addresses` instead of `/email_addresses`?

wintermeyer:
Nesting this route seems to make the most sense. This is what I as a user would expect.

riverrun:
All fixed in #615

PR #615, opened 2019-05-01
Email controller updates

### LINKED ISSUE #600: Add tests for email address controllers
Need to add comprehensive tests for email address controllers.

riverrun:
Fixes #600 and #613

### DIFF
--- a/lib/vutuv_web/controllers/email_address_controller.ex
+++ b/lib/vutuv_web/controllers/email_address_controller.ex
@@ -1,67 +1,67 @@
 defmodule VutuvWeb.EmailAddressController do
   use VutuvWeb, :controller
 
+  import VutuvWeb.Authorize
+
   alias Vutuv.{Accounts, Accounts.EmailAddress}
 
   @dialyzer {:nowarn_function, new: 2}
 
-  def index(conn, _params) do
-    user = conn.assigns.current_user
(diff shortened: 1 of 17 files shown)


lib/vutuv_web/controllers/email_address_controller.ex
```diff
@@ -1,67 +1,67 @@
...
-        |> redirect(to: Routes.email_address_path(conn, :show, email_address))
+        |> redirect(to: Routes.user_email_address_path(conn, :show, user, email_address))
 
       {:error, %Ecto.Changeset{} = changeset} ->
         render(conn, "new.html", changeset: changeset)
     end
   end
 
-  def show(conn, %{"id" => id}) do
+  def show(conn, %{"id" => id}, _user) do
```
wintermeyer, line 34: How do we ensure that a user doesn't change the `:id` part of the URL to access an email address of an other account?

wintermeyer (approved):

lib/vutuv_web/controllers/email_address_controller.ex
riverrun: There is a custom action function which checks the params and the current_user before the `show` function is called.

riverrun: In the email_controller tests, there is a test for this case, at the moment, on line 84.

Issue #621, opened 2019-05-09
Skill, UserSkill, Endorsement Scaffold


hendri-tobing:
Skill, UserSkill, Endorsement Scaffold.

Based on Vutuv V1.1

skills schema:
    field :name, :string
    field :downcase_name, :string
    field :slug, :string
    field :description, :string
    field :url, :string
    has_many :user_skills

user_skills schema:
   belongs_to :user/:profile
   belongs_to :skill
   has_many :endorsements

endorsements schema:
   belongs_to: User/Profile
   belongs_to: UserSkill

wintermeyer:
Maybe we should rename this to `tags`. A `tag` could be used for all sorts of things. One could tag a post too. The `skill` idea was in retrospect too ambitious and didn't pay out.

`endorsements` are still good.

Keep it simple. Please do not just copy v1 functionality 1:1.

wintermeyer:
And please do not copy the translation idea from v1. That was a total disaster. A good idea but for that we'd need a lot of funding to be implemented. Keep it simple and don't do translations of `tags`.

riverrun:
Fixed in #624

PR #624, opened 2019-05-19
Added tags in general context, install slug dependency

hendri-tobing:
Added tags in general context, since tags will be used by many context like user_tag, post_tag, etc.  This PR also install a slug dependency.


lib/vutuv/generals.ex
```diff
@@ -0,0 +1,84 @@
...
+  """
+  @spec get_tag(integer) :: Tag.t() | nil
+  def get_tag(id), do: Repo.get(Tag, id)
+
+  @doc """
+  Creates a tag.
+  """
+  @spec create_tag(map) :: {:ok, Tag.t()} | changeset_error
+  def create_tag(attrs \\ %{}) do
+    downcase_name =
```
riverrun: Let's call `is_binary(attrs["name"])` just once - for example:

```
{downcase_name, slug_value} = if name = is_binary(attrs["name"]) do
  {String.downcase(name), Slugger.slugify_downcase(name)}
end
```

riverrun (changes requested):
Just one change requested.

Could you also give me a little more information, or an example, about how we would use tags?

lib/vutuv/generals.ex
hendri-tobing: This is done. Thank you.

hendri-tobing:
> Just one change requested.
> 
> Could you also give me a little more information, or an example, about how we would use tags?

Initially, I thought this schema name is "skills". As Stefan advised on #621, it should be renamed as "tags" because it can be more general and also can be used for other things, for example (belongs_to) user_tag or post_tag. 

For users, user_tag can be users' skills like "Elixir", "Erlang", etc. People can endorse those user_tag.

The schema of user_tag can be like this
```
belongs_to :user
belongs_to :tag
has_many :endorsements
```

riverrun (approved):

PR #627, opened 2019-05-30
Added ProfileTag (skill) context and tests

hendri-tobing:
Added ProfileTag (skill) context and tests

ProfileTag is a many_to_many relationship between Profile and Tag. It can be used by user to put their skills on their profile page.


priv/repo/migrations/20190513223050_create_user_tags.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Vutuv.Repo.Migrations.CreateProfileTags do
```
riverrun: Can we change the name of the file to `DATE_create_profile_tags.exs`?

test/vutuv/biographies/biographies_test.exs
```diff
@@ -209,6 +207,86 @@ defmodule Vutuv.BiographiesTest do
     end
   end
 
+  describe "profile_tags" do
+    alias Vutuv.Generals
+    alias Vutuv.Biographies.ProfileTag
+
+    @valid_tag_attrs %{
+      "description" => "some description",
```
riverrun: Let's improve this description - change it to a kind of description we would expect a user to set. This can help other developers understand this code better.

riverrun (changes requested):
Just a couple of changes requested.

riverrun (approved):
Good!

Issue #625, opened 2019-05-24
Manage new email addresses


riverrun:
## Problem

As mentioned in #620, many users do not enter their email addresses correctly. This means that unless these email addresses are managed correctly, there will be many incorrect email addresses store in the database and registered to users.

## Solution

Add a process that manages email addresses: when an email address is created, this email management process will be notified that we are awaiting email confirmation. If the email address is created, the email value will be removed from this process's state. If there is no confirmation within the time limit, further action will be taken. Initially, this further action will be that the email address will be removed from the database, but in the future, we could send a reminder to the user before deleting it.

## Additional information

Although we will be implementing stricter email validation, as stated in #620 comments, there will still be incorrectly input email addresses that we will need to deal with, and that is why we will need this additional process.

wintermeyer:
In the long term it would be useful to pipe sendmail/postfix error emails from into this workflow. So when we get an error message from postfix that this email address doesn't exist we can mark it accordingly.

Unfortunately we can not mark it as unfit for ever because there is the chance that it will become an active email address in the future.

Probably deleting those uses with a cron job once a day is the best way to go.

riverrun:
Added in #628

Issue #620, opened 2019-05-06
email address format validation


wintermeyer:
It is unbelievable how many new users don't enter their email address correctly. Mostly simple typos like: "X yz <xyz@gmail>" (missing .com at the end). It would help to offer a simple email address validation which uses a regular expression. I know that a proper regular expression for email addresses is easier said than done. But an 80% solution would make a big difference here.

@riverrun Please close this issue in case you are already doing this. I just walked through some error emails and created this issue because of those errors in version 1 of vutuv.

riverrun:
Could you provide some examples of typos? Then we can test some of these.

I also need to open an issue about running a server that will delete emails that are not confirmed / verified (or that will schedule a reminder to the user before deleting them).

wintermeyer:
Believe it or not these are all real example of the last couple of days:

- iviwekulati@gmail.com2002
- victoriathankgod7@gmaill.com
- http//@bolomo.com
- frankdaud234@gmail
- mudassir.com@pk
- ricoandrew592@gmail

wintermeyer:
Just an idea for the future: We could use `<meta http-equiv="refresh" content="3; URL=https://www.vutuv.de/foobar">` to auto refresh the status page to give the user an update about his/her email. So when we get an SMTP error we could display it there. Of course not all browsers will auto refresh but again 80% is better than nothing.

riverrun:
We already check the email format on the backend, but we probably should make sure that we check it in the form as well (using the [`email` input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email) - together with `pattern`). We also need to make the email formatting regex that we are using stricter / more accurate, although according to the [emailregex site](https://emailregex.com), it might be impractical to make it 100% accurate.

Based on the pages cited above and [this page which outlines the email syntax rules](https://help.returnpath.com/hc/en-us/articles/220560587-What-are-the-rules-for-email-address-syntax-), we will use one of the following two regexes (the second one is stricter with regard to the top-level domain):

```
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9-\.]+\.[a-zA-Z0-9-]+$/
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9-\.]+\.[a-zA-Z]{2,}$/
```

riverrun:
Added in #628

Issue #614, opened 2019-04-30
No JavaScript


wintermeyer:
The version 1.0 doesn't use JavaScript. This is mainly because of https://www.internet.org/ which delivers vutuv to it's users for free. This is actually a big deal because LinkedIn is not on that platform and internet.org delivers internet to many people who wouldn't have it otherwise. We do have quite a few vutuv users from internet.org

In version 2.x I'd like to integrate a non JavaScript version and a version which uses JavaScript to make a nicer GUI. But because we can not create both worlds in the 2.0 version we have to go with the non JavaScript version and implement the JavaScript version for 2.1.

We can use cookies. Just not JavaScript.

I remember that in version 1.0 this was easier said than done because the default Phoenix scaffolding uses JavaScript.

@riverrun and @hendri-tobing: Please share your thoughts.

hendri-tobing:
The purpose of a non-JavaScript version is for the fast performance. The purpose of the JavaScript version is for a nicer GUI. Which one will give us more benefits/users? I agree in version 2.1, we can use JavaScript as optimum as possible to allow a nice GUI and also relatively still fast application.

riverrun:
The only javascript currently being used is from the Phoenix HTML application, and it is used with forms.

When removing this javascript, delete actions might not work properly, and there might be other issues related to form handling.

riverrun:
The problem with (and solution for) using the delete method in forms is outlined in [this SO question](https://stackoverflow.com/questions/28164132/how-to-patch-or-delete-from-a-form-without-javascript).

@wintermeyer do you know if this kind of workaround works with https://www.internet.org/?

wintermeyer:
> The problem with (and solution for) using the delete method in forms is outlined in [this SO question](https://stackoverflow.com/questions/28164132/how-to-patch-or-delete-from-a-form-without-javascript).
> 
> @wintermeyer do you know if this kind of workaround works with https://www.internet.org/?

I do not know. But `<input name="_method" type="hidden" value="patch" />` seems to be really simple. This should AFAIK work. But please have a look into version 1 to see who it was solved in it.

riverrun:
@wintermeyer what is the best commit to look at for version 1? I looked at the [commit before the initial work was started for v2](https://github.com/vutuv/vutuv/commit/41fcb9e87da67c12adde3671d9d40d78a0f339cc). Is this the best one?

wintermeyer:
https://github.com/vutuv/vutuv/tree/v1.1

[...]

Issue #610, opened 2019-04-18
Temporary CSS for Profile Page


hendri-tobing:
While waiting issue #576 is resolved, I will try to reuse the CSS from Vutuv-v1 for the Profile Page, temporarily.

riverrun:
Before switching to Vutuv-v1 css, I am going to use the default phoenix css, as this is quite lightweight and should 'just work'.

riverrun:
Added in #628

PR #628, opened 2019-06-04
Email management and simplification of css

riverrun:
This PR adds email management (for removing invalid / unverified email addresses) and updates email format verification, both on the frontend and backend - issues #625 and #620.

It also changes the css from the v2 version to the default Phoenix css and removes the javascript from the app - issues #610 and #614.


lib/vutuv_web/templates/email_address/form.html.eex
```diff
@@ -1,7 +1,7 @@
 <%= form_for @changeset, @action, [], fn f -> %>
   <%= if @changeset.action do %>
     <div class="alert alert-danger">
-      <p>Oops, something went wrong! Please check the errors below.</p>
+      <p>Something went wrong. Please check the errors below.</p>
```
wintermeyer, line 4: How do we handle I18n of this message?

lib/vutuv_web/templates/profile/form.html.eex
```diff
@@ -1,82 +1,59 @@
...
+  <%= text_input f, :honorific_suffix %>
+  <%= error_tag f, :honorific_suffix %>
 
-  <div class="form-group">
-    <%= label f, :gender, class: "sr-only" %>
-    <%= select f, :gender, ["Gender": "", "Female": "female", "Male": "male", "Other": "other"], [class: "form-control"] %>
-    <%= error_tag f, :gender %>
-  </div>
+  <%= label f, :gender %>
+  <%= select f, :gender, ["Gender": "", "Female": "female", "Male": "male", "Other": "other"] %>
```
wintermeyer, line 35: How do we handle I18n of these values?

riverrun: I think it would be better to open a separate issue about that, and we can go through all the templates and make sure the I18n works for them.

wintermeyer (approved):

Issue #630, opened 2019-06-05
Title case or sentence case


riverrun:
While working on #629, I noticed that some headings are written with all capital letters (title case) and some are written with just the first word capitalized (sentence case).

I think it would be useful to decide which one we want to use and stick with it. That will make the site look more professional.

Personally, I don't have a strong opinion on this matter, but I do prefer sentence case.

More information about the differences can be found [Why letter casing is important](https://uxplanet.org/why-letter-casing-is-important-to-consider-during-design-decisions-50402acd0a4e) and [Making a case for letter case](https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98).

wintermeyer:
Good catch! Let's go with sentence case.

riverrun:
The templates now use sentence case. I need to add a short note about this to the CONTRIBUTING guide, and then we can close this issue.

riverrun:
Fixed in #634

PR #634, opened 2019-06-07
Enabled delete capability from forms - without javascript

riverrun:
Fixes 'delete' problem mentioned in #614 and adds comment about #630 to the CONTRIBUTING guide.

### DIFF
--- a/lib/vutuv_web/templates/email_address/index.html.eex
+++ b/lib/vutuv_web/templates/email_address/index.html.eex
@@ -26,7 +26,11 @@
       <td>
         <%= link gettext("Show"), to: Routes.user_email_address_path(@conn, :show, @current_user, email_address) %>
         <%= link gettext("Edit"), to: Routes.user_email_address_path(@conn, :edit, @current_user, email_address) %>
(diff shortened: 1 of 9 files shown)


lib/vutuv_web/templates/email_address/index.html.eex
```diff
@@ -26,7 +26,11 @@
       <td>
         <%= link gettext("Show"), to: Routes.user_email_address_path(@conn, :show, @current_user, email_address) %>
         <%= link gettext("Edit"), to: Routes.user_email_address_path(@conn, :edit, @current_user, email_address) %>
-        <%= link gettext("Delete"), to: Routes.user_email_address_path(@conn, :delete, @current_user, email_address), method: :delete %>
+        <li>
+          <%= form_for @conn, Routes.user_email_address_path(@conn, :delete, @current_user, email_address), [method: "delete"], fn _f -> %>
+          <%= submit gettext("Delete"), class: "link" %>
```
wintermeyer, line 31: The indentation seems to be off. But I don't know if we do strict indentation in eex at all. Never mind in case everything is all right here.

wintermeyer (approved):

Issue #633, opened 2019-06-06
Name fields - needs and uses


riverrun:
According to [this article about making more inclusive forms](https://www.uxbooth.com/articles/creating-more-inclusive-and-culturally-sensitive-forms/) and [this article which argues for only having one name field](https://uxmovement.com/forms/why-your-form-only-needs-one-name-field/), many users are left out when asked to fill in forms with the traditional 'First name' and 'Last name' fields. In addition, we have 3 fields for the name, and I think the users will find it easier if we just offer them one - a 'Full name' field.

Also, we are asking users for their 'Nickname', which seems a little too informal. I think we should consider using 'Preferred name' instead. Of course, this does also depend on how this information is used, but I think that 'Preferred name' is more suitable.

Any thoughts?

panomaki:
I think there are a couple of issues to be solved here which have to do with the usage of the field. The first of which is that if it is used to find people, it should be designed to allow for all ways in which people are known, including first name, last name and nickname, This *could* be saved in one field, if the search algorithm is designed appropriately. Second is the way in which people would like to be addressed, which may or may not be related to what is saved in the name-field mentioned earlier. This may be a more formal addressation - Dr Wossname, Daniel-San, Weledelgestrenge heer Woudstra, etc. There may be possibilities in localization/i8n environments to address this, but it's not my field of knowledge. It may also be an informal addressation - nickname, only firstname, or whatever. So two more fields - ah, you have three already ;-). One for name, one for formal addressation, one for informal addressation?

Just 2 cents for the discussion...

wintermeyer:
**First of all: Great issue!** That's the kind of discussion I really enjoy. I think that we have to do a better job with this than LinkedIn or XING. That said: I've been thinking about this for the last 20 something hours and haven't found a one-size-fits-all answer.

The reason why we started with `first_name`, `last_name` and `gender` is that this gives us the possibility to create a proper greeting in an email. In German an email would start with "Hallo Herr Wintermeyer," or "Sehr geehrter Herr Wintermeyer," (more formal) but a "Hallo wintermeyer," (account name) would be rude. A "Hallo Stefan Wintermeyer," shouts out loud "we didn't know what gender you are and therefor we do this". A "Hallo Stefan," doesn't go well in German either because that would be a little too less formal (at least for the elder generation).

So for Germany and many other countries this would be the best way to go. But I realize that the world is bigger and that we want to be as inclusive as possible. 

[...]

PR #635, opened 2019-06-07
Simplify forms

riverrun:
This PR mainly reduces the number of inputs in the forms to make them more user-friendly.

The birthday entries have been put into one date_of_birth field, the names have been combined into a full_name field and nickname has been replaced with preferred_name (see #633 for more information about the name changes).

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -37,8 +37,6 @@ config :phoenix, :json_library, Jason
 config :arc,
   storage: Arc.Storage.Local
 
-config :slugger, separator_char: ?-
-
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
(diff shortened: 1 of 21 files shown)


wintermeyer (approved):
The original idea was that somebody could be willing to enter his/her birthday but without the year to not show how old he/she is. But from todays perspective that was overly complicated. If somebody doesn't want to share his/her age that he/she shouldn't enter the DOB at all.

riverrun:
> The original idea was that somebody could be willing to enter his/her birthday but without the year to not show how old he/she is. But from todays perspective that was overly complicated. If somebody doesn't want to share his/her age that he/she shouldn't enter the DOB at all.

That's good to know. I might be able to add that kind of functionality. I'll see what I can do.

Issue #558, opened 2019-02-21
New Sign Up Page


wintermeyer:
# Introduction

vutuv is a business network. Think of it as a free, fast and secure open-source alternative to [LinkedIn](https://www.linkedin.com) or [XING](https://www.xing.com). Please feel free to add comments and questions (as comments) to this issue.

# The Task

**Create a mockup sign up HTML page.** English or German (language doesn't matter for this mockup). To solve this issue no Elixir code has to be written. Just plain old HTML and CSS (probably embedded in the HTML). If you need data use dummy data. Please store the file(s) in `misc/mockups/` or a subdirectory of it. The HTML file should contain the form. Later on someone will convert this mockup into a Phoenix template. But that will be a separate issue. The old version of this sign up page is attached to this issue in a comment.

Please create a pull request to submit your work (master branch).

# General Thoughts about the Sign Up Screen

If you visit https://www.vutuv.de without being logged in this is your first page. I still like the idea and the feeling of the old one (in the first comment you'll find a gz of it). It would be nice to keep the idea but we don't have to.

<img width="1228" alt="The old login screen." src="https://user-images.githubusercontent.com/69770/53193838-4b2c0c80-3612-11e9-8a4d-afb4bd9e431b.png">

**The sign up screen is the first vutuv page new users see.** It has to be positive and welcoming. If we can exchange the image during the day or the month it would be nice. But that is not a necessity. 

## WebPerformance

Speed is paramount. Our payload budget for HTML, CSS, the SVG embedded vutuv logo and JavaScript (if needed) for this page is 14 KB (gzip compressed). We load one image (the background image) which is not included in the 14 KB.

### JavaScript

I think we don't need JavaScript on the sign up page. But I'm open for it in case it gives us big benefits.

### Font

Only use system fonts. No fonts which have to be loaded.

### The Background Image

The image should be taken by one of us or from a free source (e.g. [unsplash](https://unsplash.com)). The credit information should be shown somewhere on the page.

The image should be not bigger than 28 KB. If we absolutely must have a bigger image the worst case limit for that image is 47 KB. But there has to be a good reason for that.

Obviously it is not easy to find good full page images which can get compressed to 28 KB or 47 KB. If it was easy anybody could do it.

### Logo

This should be an inlined vector image.

### Misc

All other buttons or graphics probably should be inlined vectors too. But I'm open for better ideas if available.
 
## Navigation

Can be a classic top navigation or something else. It has to have these two elements: 

- A login link
[...]

PR #637, opened 2019-06-12
Improve css

riverrun:
This updates the css to look like the version outlined in #558, but the css is kept lightweight.


wintermeyer (dismissed):

riverrun:
I just updated the travis file - let's see if it works now.

wintermeyer (approved):

PR #649, opened 2019-07-01
Add locale

### LINKED ISSUE #646: locale
Currently the `locale` of a `user` is set to `nil`. In v1 we used the HTTP header to fetch the locale. Of course this could be wrong but in 95% of all cases it is a very good value for this. But users have to be able to override it.

Problem: We do not support all languages. In the beginning just `de` and `en`. We have to find a mapping where `en` is the default language if the `locale` is not yet supported.

riverrun:
Fixes #646. Also adds the vutuv svg logo to the background for large screens.

The locale is now added when a new user is created. The locale generation parses the "accept-language" header and tries to find the supported language with the highest quality. If no supported language is found, the locale is set to "en".

Supported locales are the English and German variants.

There is also an option for the user to update the locale. When updating the locale, the update operation is only successful if the new locale is supported - if it is not supported, then no change is made.

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -58,9 +58,11 @@ body {
   background-color: #4a6a90;
   color: #e6e6e6;
   display: grid;
-  grid-gap: 0;
-  grid-template-rows: auto 1fr auto;
   grid-template-columns: 1fr;
+  grid-template-rows: auto 1fr auto;
+  grid-template-areas: "header"
+                        "main"
+                        "footer";
 }
 header,
 main {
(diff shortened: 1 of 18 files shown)


wintermeyer (changes requested):
Let's store the original `accept-language` value in the database too. No need to display that value to the user but it gives us the possibility to see what languages are popular so that we can do new I18n work depending on that data.

riverrun:
Now accept-language header entry is added to the users table when a new user is created. This header is also used to generate the locale.

The accept-language is added on user creation and cannot be updated.

riverrun:
Superseded by #650

Issue #643, opened 2019-06-18
Add more finely-grained authorization for email_addresses


riverrun:
We need to make sure that certain email_address data are not leaked to unauthorized users. For example, if an email_address is meant to be private, it should not be displayed to other users.

In this case, we will handle this authorization in the accounts context module. If the current user is the owner of the data, then the full email_address data can be show, but if the current user is someone else or anonymous, only publicly available data will be shown.

riverrun:
Closing this - the necessary context functions were added in #652.

PR #652, opened 2019-07-09
User credentials

### LINKED ISSUE #651: Reorganize directory / table structure with regard to user data and profile data
Currently, we have user data split up into user account data (users) and user profile data (profile). As mentioned by @hendri-tobing (in #579), this separation is recommended at [cloud google](https://cloud.google.com/blog/products/gcp/12-best-practices-for-user-account). It is also recommended at [swoopnow](https://swoopnow.com/web-application-authentication-best-practices/), which provides more information about this subject.

## Problem

The main problem, as I see it, is that we are not enforcing the separation strictly / clearly enough. When we load user data, for example, we are also including user_credential data. This can be solved by adjusting how we query the data, but I prefer to reorganize the table structure, as that should make this policy of separating user data from user account / credential data clearer.

A smaller problem is that in the current setup there is less of a clear distinction between modules that handle user data and those that handle profile data. For example, the accounts module has several functions that also handle profile data, which makes the functions in biographies (profile) context redundant.

## Solution

What I propose here is that we reorganize how we separate the data, as shown below:

* move profile table to users table
* move the old users data to user_credentials
* move profile_tags to user_tags
* move phone_number to belong to users

Along with these changes, I suggest that we only load the user (user profile) data when setting the current_user. At the moment, we load user data, user_credential data and email_address data - we definitely should not need user_credential data, and I think email_address data is not really needed in most cases.

## Further information

Additional information can be found at the following links:

https://softwareengineering.stackexchange.com/questions/241089/keep-user-and-user-profile-in-different-tables
https://stackoverflow.com/questions/3137080/table-design-for-users-information-as-well-as-login-credentials
https://blog.goyello.com/2012/01/10/not-store-user-credentials-database/
https://softwareengineering.stackexchange.com/questions/241089/keep-user-and-user-profile-in-different-tables
https://stackoverflow.com/questions/3889797/mysql-user-profile-details-table-setup-best-practice
https://stackoverflow.com/questions/36028538/does-it-make-sense-to-separate-the-table-of-user-profile-to-that-of-the-login-de

riverrun:
Fixes #651 and #643.


lib/vutuv_web/templates/user/edit_form.html.eex
```diff
@@ -5,62 +5,60 @@
...
-        {gettext("August"), "8"},
-        {gettext("September"), "9"},
-        {gettext("October"), "10"},
-        {gettext("November"), "11"},
-        {gettext("December"), "12"},
-      ]
-    ] %>
-    <%= error_tag fp, :birthday %>
+  <%= label f, :birthday, gettext("Date of birth") %>
+  <%= date_select f, :birthday, year: [options: 1900..2020, selected: 1980], month: [
```
wintermeyer: The 2020 has do be done dynamically being always the current year.

wintermeyer (changes requested):
Just one small change.

riverrun:
Fixed that issue, and made a small update to the user create_changeset (to make it a bit stricter).

wintermeyer (approved):

Issue #654, opened 2019-07-10
Make sure only accessible links are present on each page


riverrun:
We should make sure that any link displayed on a page should be accessible by the current user.

At the moment, for example, if an anonymous user visits a user's page, there will be links to edit the user, or edit email addresses, even though the anonymous user will not be able to go to these pages. We should remove these links depending on who the current user is.

riverrun:
Fixed in #655

PR #655, opened 2019-07-12
Update links

### LINKED ISSUE #611: Use Task to get the gravatar image
Currently, the function to get the gravatar image is a blocking call which will stop the user being created until the image has been downloaded. We should use an async call - using Task.async - to improve the responsiveness of the `create_user` function call.

riverrun:
Fixes #611 and #654.

## Gravatar downloads

This PR adds a separate process that downloads gravatar images asynchronously after a user has been created, and if needed, updates the database with the user avatar. If there is any error, the download process is not retried (and the default avatar will be used).


lib/vutuv_web/controllers/user_controller.ex
```diff
@@ -33,6 +33,7 @@ defmodule VutuvWeb.UserController do
       {:ok, user} ->
         Log.info(%Log{user: user.id, message: "user created"})
         email = user_params["email"]
+        # maybe move these lines to confirm/new
```
wintermeyer, line 36: @riverrun Did you forget to remove this comment or is this on purpose?

wintermeyer (approved):

lib/vutuv_web/controllers/user_controller.ex
riverrun: I forgot - I will remove it next commit.

Issue #577, opened 2019-03-04
Posts


wintermeyer:
# The Idea

More text than Twitter but less than Medium. With markdown as the language of choice to write articles and the possibility to add photos. 

## Different Levels of Length and Features

- New users can only write 140 character posts with no markdown (just auto linking URLs). 
- Next level users can write 280 characters.
- Next level users can use markdown features.
- Next level users can write 512 characters.
- Next level users can write 2048 characters.

The newest posts will show up on the profile page (at least an excerpt if it is a longer post) and old posts will be available at `/users/:slug/posts` which is an index page with pagination. Each single post will be available at `/users/:slug/posts/:id`

## Twitter Cards

If a post contains a URL than essential information of that page should be displayed below the post in the profile list (e.g. Twitter card info or HTML title if Twitter card is not available).

## Tags

A post has tags (e.g. PHP, Origami, Travel).

# Edit/Delete

Yes, posts can be edited after being submitted. And of course they can be deleted.

# Attributes

- :title, :string
- :body, :string
- :page_info_cache, :string # to store the content of Twitter cards, etc.
- :user_id, :integer
- :publicated_at # Timestamp from when on the post is published.
- :visibility_level, :string # Drop Down Select [public|just for followers|private]

riverrun:
Working on this and finished the basic outline of it.

One question: what do we want the default `visibility_level` to be - public, private or just_followers?

wintermeyer:
> One question: what do we want the default `visibility_level` to be - public, private or just_followers?

private

wintermeyer:
Thinking of it: If it is set to private how can a potential new follower follow it?

riverrun:
The 'Different Levels of Length and Features' should not be implemented at the moment. This feature is under review. For more information, see [this issue comment](https://github.com/vutuv/vutuv/pull/660#issuecomment-512119965).

riverrun:
Mostly done in #660.

Still to do are the Twitter cards / HTML tite and restricting access to just a user's followers.

riverrun:
Closing this issue - added separate issue regarding Twitter cards / HTML title.

PR #660, opened 2019-07-17
Posts

riverrun:
This PR adds posts - providing most of the functionality outlined in #577.

## Further details

Added:

* post routes
    * these are nested under user
    * the index and show routes are open to any user (including anonymous users), and all other routes can only be accessed by the owner of the post
* post tags

Still todo:

* twitter cards
* different capabilities for different levels of users (for example, new users can only write 140 chars)
* limiting posts to just followers (followers not implemented yet)


wintermeyer:
Please don't implement `different capabilities for different levels of users (for example, new users can only write 140 chars)`. I'm not sure about this any more and we don't need it in the first 2.0 version. So no limits about the length of a post. OK. Maybe a very big limit just so that no-one can crash the DB.

riverrun:
Ok. I will add a note to the issue.

For the body of each post, the current limit is set to 255 chars.

wintermeyer:
> For the body of each post, the current limit is set to 255 chars.

That should be a lot more. More like 100,000 chars. Or is there a limitation on the DB side for such a big text? People should be able to post complete book chapters if they feel like it.

riverrun:
Ok. Let me look into our options - how big we can / should be making it - and then I can update it.

One other I should have mentioned is that markdown is not yet supported.

riverrun:
According to [this guide](https://blog.hubspot.com/marketing/character-count-guide), Facebook set a limit of 63,206 characters, and Linkedin set a limit of 40,000 characters for blog posts.

In our case, shall we make it around 50,000, 60,000, or more?

wintermeyer:
> According to [this guide](https://blog.hubspot.com/marketing/character-count-guide), Facebook set a limit of 63,206 characters, and Linkedin set a limit of 40,000 characters for blog posts.
> 
> In our case, shall we make it around 50,000, 60,000, or more?

Let's make it 126,412 (double the amount of Facebook). No, forget that. Funny but doesn't make sense. Let's do 150,000 which is an easy to remember number and give plenty of room for everybody to write stories.

wintermeyer (approved):

PR #669, opened 2019-08-07
Update error handling

### LINKED ISSUE #668: Handle not found (404) errors using Ecto.NoResultsError
Currently, the 404 errors are being handled manually. There should be no need for this, and it will simplify the implementation if we use Phoenix's default handling of Ecto.NoResultsError - which then sets the status to 404.

riverrun:
Fixes #668 and also makes some minor changes to how email verification is handled, adding a `send_code` route so that users can have the code resent to them.


riverrun:
Closing this PR - superceded by #670.

Issue #675, opened 2019-08-14
Need to implement for version 2.0


riverrun:
The following models are in version 1.1, but have not been implemented in version 2.0 yet.

* address (physical address)
* group
* job_posting
* job_posting_tag
* membership
* recruiter_package
* recruiter_subscription
* search_query
* social_media_account
* user_tag_endorsement
* work_experience

Are all of these necessary? Is there anything else that needs to be added to this list?

wintermeyer:
What we need:

- address (physical address)
- social_media_account
- user_tag_endorsement (some kind of)
- work_experience

riverrun:
Quick update: working on address and social_media_account now.

PR #678, opened 2019-08-21
Added work_experience table and routes

riverrun:
As part of #675, added work_experience table and routes.


wintermeyer (approved):
With the amount of pain killers I've taken this week I don't expect this to be the most thoroughly done review ever.

PR #754, opened 2026-06-02
Remove Google+ from the base URI list.

### LINKED ISSUE #741: Replace Google+ with Fediverse
Sadly G+ has long gone, but maybe it can be replaced with Fediverse, allowing the addition of any Fediverse address (same ActivityPub protocol) from Mastodon, Friendica, Hubzilla, PixelFed, etc.

beedell-roke:
Removed Google+ from the list of base URIs for social media accounts. This is the closest resolution to [`issues/741`](https://github.com/wintermeyer/vutuv/issues/741#event-26239213898) that I am able to provide.

#### Linkage

Resolves [`issues/741`](https://github.com/wintermeyer/vutuv/issues/741#event-26239213898).

### DIFF
--- a/lib/vutuv/profiles/social_media_account.ex
+++ b/lib/vutuv/profiles/social_media_account.ex
@@ -21,7 +21,6 @@ defmodule Vutuv.Profiles.SocialMediaAccount do
   base_urls = [
     {"Facebook", "http://facebook.com/"},
     {"Twitter", "http://twitter.com/"},
-    {"Google+", "https://plus.google.com/+"},
     {"Instagram", "http://instagram.com/"},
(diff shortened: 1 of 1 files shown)


wintermeyer:
Thank you for spotting this and taking the time to send a fix, @RokeJulianLockhart — it's much appreciated! 🙏

I'm closing this in favour of #755, which builds on your change. Google+ was referenced in four places (changeset validation, link generation, display prefix, and the form `<select>`), so removing only the base URL entry left it half-wired: still selectable and accepted, but producing an empty link. #755 removes all four and adds a test to cover it.

Credit for catching the issue is yours — thanks again for contributing!

PR #1120, opened 2026-07-26
Quote a marked passage into a reply (#1114)

### LINKED ISSUE #1114: Feature Request: Quote certain sections from posts
# Motivation
Especially when answering longer posts, it would be convenient, to just select the paragraph or sentence you want to quote, have a little context menu there, which offers to quote the selection. The selection would then result in a block quote, but the block quote would render with the avatar of the user you are quoting as well as the link to the post, you are quoting.

wintermeyer:
**TL;DR** Mark part of a post, press that post's **Reply**, and the composer opens with the marked text already in it as a blockquote. Closes #1114.

**Where:** `assets/js/app.js`, `VutuvWeb.Markdown.blockquote/1`, `VutuvWeb.PostLive.Reply`, `VutuvWeb.PostComponents.post_card/1`
**Now:** answering one point of a long post means retyping it by hand.
**Want:** select the sentence, hit Reply, write the answer under the quote.

## How it works

No floating menu, no new route, nothing stored. The selection rides on the Reply link the card already has:

```
select text in a post  ->  click Reply  ->  /posts/<id>/reply?quote=<text>
                                             |
                                             VutuvWeb.Markdown.blockquote/1
                                             |
                                             composer opens with "> …"
```

`app.js` reads the selection at click time (the browser collapses it on `pointerdown`, so the pointerdown pass remembers it; a keyboard activation still has the live one). The reply page runs that untrusted string through `VutuvWeb.Markdown.blockquote/1`: trim each line, keep the quote marker across blank lines so a paragraph break stays inside the quote, cap at 500 characters cut back to the last whole word.

The server marks the two halves the client needs: `data-post-body` on a post's prose and `data-quote-reply` on the enclosing card, naming its Reply control by id. Thread cards nest, so the card is resolved from the **selection** (innermost marked ancestor), not from the link. Otherwise a marked reply would quote itself into its parent's answer.

Two places deliberately stay unarmed: a **restricted post** (its Reply control is a dead span) and the reply page's **own parent preview**, whose Reply link leads back to the page being composed on and would throw the draft away.

## About the avatar and the link

The issue asks the blockquote to carry the quoted member's avatar and a link to the post. It already does, one level up: the reply card names the post it answers, with that author's avatar and the thread connector into it. So the attribution needs no per-blockquote metadata and the stored body stays plain Markdown, which keeps the agent-format siblings, the fediverse copy and the RSS rendering untouched.

## Editor bug fixed on the way

Found smoke-testing this: a document whose last node is a blockquote leaves ProseMirror nowhere to put the cursor, so clicking below the quote typed the answer **into** the quotation. The Milkdown hook now appends an empty trailing paragraph after every seed, which also unsticks a body ending in a table or a code block. It costs nothing in the stored source (Milkdown serializes it as the `<br />` artifact `normalizeMarkdown` already drops).

## Verification

`mix precommit` green (5255 tests). New coverage: `markdown_quote_test.exs` for the excerpt shapes a selection really has, plus reply-page and feed-card tests for the seeding and the two unarmed cases.

Smoke-tested in a real browser, since the whole trigger is client-side and no ConnTest can reach it: real drag selection on a post permalink, real click on Reply, quote in the URL, blockquote in the WYSIWYG editor, answer typed below it, reply posted and rendered under its parent. Also verified the negative case, a selection inside the nested reply followed by a click on the **parent's** Reply, which correctly carries no quote.

## Not in scope

A reply that came from another network (`remote_reply_card`) is not armed yet. Same three lines if we want it, but it is a different card and its Reply link is a live navigation, so I kept it out of this change.

https://claude.ai/code/session_01H6dRXwvHciTBK8fd9CNJSu

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -160,6 +160,74 @@ window.addEventListener("resize", () => {
   previewClampResizeTimer = setTimeout(sweepPreviewClamps, 150)
 })
 
+// Quote a passage into a reply (issue #1114). Mark part of a post, press that
+// post's Reply control, and the reply page opens with the marked text already
+// in the composer as a blockquote. No new chrome: the selection rides along on
+// the Reply link the card already has, as a `quote` query parameter that
+// VutuvWeb.Markdown.blockquote/1 normalizes and caps server-side.
+//
+// The server marks the two halves — `data-post-body` on a post's prose, and
+// `data-quote-reply` on the enclosing card naming its Reply control. Thread
+// cards nest, so resolving the card from the SELECTION (innermost marked
+// ancestor) rather than from the link is what keeps a marked reply from
+// quoting itself into its parent's answer.
+const QUOTE_MAX = 500
+
+// What was selected when the pointer went down. The browser collapses a
+// selection as part of the default action of pointerdown, i.e. before the
+// link's click event ever fires, so by click time it is already gone.
+let pendingQuote = null
+
+function selectionQuote() {
+  const selection = window.getSelection()
+  if (!selection || selection.isCollapsed) return null
+
+  const text = selection.toString().trim()
+  if (!text) return null
+
+  const anchor = selection.anchorNode
+  const start = anchor && (anchor.nodeType === 1 ? anchor : anchor.parentElement)
+  const body = start && start.closest("[data-post-body]")
+  const card = body && body.closest("[data-quote-reply]")
+  if (!card) return null
+
+  return { text: text.slice(0, QUOTE_MAX), replyId: card.dataset.quoteReply }
+}
+
+// Capture, so a handler that stops propagation can't rob us of the selection.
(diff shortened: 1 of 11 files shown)


shaedrich:
> ## About the avatar and the link
> 
> The issue asks the blockquote to carry the quoted member's avatar and a link to the post. It already does, one level up: the reply card names the post it answers, with that author's avatar and the thread connector into it. So the attribution needs no per-blockquote metadata and the stored body stays plain Markdown, which keeps the agent-format siblings, the fediverse copy and the RSS rendering untouched.

@wintermeyer This way, it is impossible to quote from different posts. Imagine, you take part not in a 1 on 1, but a multi-user conversation, come back the next day, find replies from various and want to address everyone in a single post. This would get confusing pretty fast if not for that exact feature.

PR #1127, opened 2026-07-26
Bump to 7.164.0 after a version collision on main

wintermeyer:
**TL;DR** Two commits on `main` both carry 7.163.0. Bumping so the version stays monotonic.

**Where:** `mix.exs`
**Now:** `a4a03ff2` (#1122's own collision fix) and `092430d9` (reply quoting, #1120) are both 7.163.0.
**Want:** the next commit on `main` reports 7.164.0.

The quote branch was rebased and re-bumped correctly off `origin/main`, but #1122 merged during its four-minute CI run, and an identical `version:` line merges without a conflict, so neither git nor CI saw the clash. Third time today, so I have sharpened my own rule to re-check `origin/main`'s version immediately before pressing merge, not only at rebase time.

Same one-line fix as #1119 and #1122.

*An AI agent wrote this text in my name, unreviewed by me. The work behind it is mine; I only delegated the writing.*

https://claude.ai/code/session_01H6dRXwvHciTBK8fd9CNJSu

PR #1129, opened 2026-07-26
Re-check main's version at merge time, not only when bumping

wintermeyer:
The version-bump rule in `CLAUDE.md` already said to compute the bump from the current `origin/main`. I did exactly that on #1121 and still collided: #1119 merged while my PR sat in CI and took 7.162.0. Because both sides then read the same number there was no merge conflict and no warning of any kind — the squash landed the whole feature without moving the version, and 7.162.0 ended up naming two unrelated changes (fixed by #1122).

That is the fourth time, and it keeps happening because the rule only covers the moment you bump, not the window between pushing and merging. On this repo that window is minutes long and usually has other PRs in it.

The rule now names that window, gives the check to run right before clicking merge (`git show origin/main:mix.exs | grep -m1 version` must still show the number you bumped *from*), and gives the after-the-fact tell: your merge commit must appear in `git log --oneline origin/main -- mix.exs`.

Docs only, so patch. Numbered 7.163.2 deliberately rather than racing #1127 for 7.164.0 — a later 7.164.0 stays monotonic on top of it.

https://claude.ai/code/session_01BvQXug8GXXiPyfrUV9z7Pr


wintermeyer:
Parking this as a draft for now, so nobody merges it in the state it is in.

The branch on GitHub still carries 7.163.2, which was correct when I pushed it but is now *behind* main (7.164.0 landed during the CI run). I have the fix ready locally, rebased onto main and bumped to 7.164.1 with `mix precommit` green, but I cannot push it: the pre-push hook re-runs the suite and `Vutuv.Repo.UpgradePathMigrationTest` gets killed by Postgres `too_many_connections` every time, because several worktrees are running their suites against the same local database. The test passes on its own and in my own full precommit run, so this is local resource contention, not the change.

Happy to push it the moment the machine is quieter. The content is a one-paragraph clarification to the version-bump rule in `CLAUDE.md`, so there is no hurry.

Issue #1104, opened 2026-07-25
Photo posts: bento galleries in the feed, original download and camera info on the permalink


wintermeyer:
**TL;DR** Photographers are business people and their photos are their work: let a post present them properly. One photo or an auto-laid-out bento mosaic in the feed, opt-in extras on the permalink (per-photo original download and camera info, a per-post license). Guiding principle: easiness beats a forest of features; everything beyond drop-photos-and-post is one switch or one select.

**Why:** vutuv stays a business network at its core; this is not a pivot to photo sharing. For photographers, designers and other visual professionals the photos *are* the portfolio, so presenting them well is profile substance, and the license select plus original download are working tools (a client fetches the deliverable straight from the post). A business network is also allowed to be human: the occasional hiking photo does the same job as small talk at a trade-fair booth. No separate photo section, no new page, just the normal feed showing images better.

**Where:** composer (`VutuvWeb.PostLive.Composer`), feed cards (`VutuvWeb.PostComponents`), permalink (`VutuvWeb.PostLive.Thread`), image pipeline (`Vutuv.PostImageStore`, `Vutuv.Uploads.Spec`, `Vutuv.Posts.PostImage`)
**Now:** a post attaches up to 10 images (6 MB each, JPEG/PNG/WebP, HEIC when decodable). Unreferenced images render as a flat gallery under the text. Served versions are metadata-stripped AVIF (`thumb`/`feed`/`large`); the original keeps its metadata in the private `originals/` tree and is never served. No layout control, no downloads, no camera info.

**Want**

*Feed (blends in, no separate post type):*
- 1 photo: full-width card, as today.
- 2+ photos: an aspect-aware bento mosaic, laid out automatically from the photos' aspect ratios. The first photo is the hero tile, card height capped, a "+N" tile when more than ~5. Clicking opens the permalink.
- Rearranging is just drag-to-reorder in the composer; the mosaic re-flows from the new order and the first photo stays the hero. No tile-size editor.

```
 3 photos             5 photos
┌────────┬────┐      ┌────────────┬────┐
│        │    │      │            │    │
│  hero  ├────┤      │    hero    ├────┤
│        │    │      │            │    │
└────────┴────┘      ├─────┬──────┴────┤
                     │     │           │
                     └─────┴───────────┘
```

*Permalink (the show, where the extras live):*
- Lightbox with full-quality versions (`large`, possibly a new `xl` spec) and per-photo captions (a `caption` field next to the existing `alt`).
[...]

PR #1139, opened 2026-07-26
German labels for the photo-post UI (#1104)

wintermeyer:
The photo feature shipped **with no translations at all**. `mix gettext.extract`
was never run for it, so on a German site the composer's photo strip, the
per-photo panel, the lightbox, the licence select and the "not public yet"
banner all rendered in English. vutuv is a German site, so that is nearly every
real visitor.

Stefan found it by asking how the feature actually works — and the answer was
partly "you can't tell, because the labels are in the wrong language".

## What is in here

45 strings translated (Sie-Form throughout). Besides the new photo UI, the guard
below turned up older gaps on the same surfaces, so those are done too: the
composer's error messages, the audience sheet's hint, the pinned-post
confirmation.

## Two traps worth knowing

Both produce **wrong German** rather than missing German, and neither is visible
when skimming a diff:

1. **`mix gettext.merge` fuzzy-guesses.** It fills a new msgid's translation
   from a similar existing one and marks it `#, fuzzy`. It had filled
   `"Remove photo"` with the German for *Remove logo*, and
   `"Show camera settings"` with *Search settings*. To anyone scanning the file
   those look translated.

2. **A fuzzy guess can be a multi-line msgstr** — `msgstr ""` followed by
   continuation fragments that gettext concatenates. Replacing only the `msgstr`
   line leaves the fragments behind, and they glue themselves onto the new text.
   That briefly produced a banner carrying two unrelated sentences at once
   ("Bisher sehen nur Sie diesen Beitrag." plus the whole of a different
   message). This is the same shape as the `.po` duplication that once 500ed the
   German landing page.

## The guard

`german_catalog_test.exs` fails the build when a string on the post composer,
the post card or the licence vocabulary is missing from the German catalog, is
untranslated, or is left `fuzzy`.

Scoped to those surfaces deliberately: the catalog carries a long tail of
untranslated legacy prose, and a test that failed on all of it would simply get
deleted. The moduledoc says to widen the list as surfaces are cleaned up, never
to narrow it to make a red build green.

## Also

Removed the em-dashes the licence labels and the progress banner had carried
into UI copy, which the project style rules exclude. That changes four msgids,
which is why they re-extract.

## Verified

`mix precommit` green (5453 tests). Checked at runtime after a clean recompile
that each string really resolves — `Photo options` → `Foto-Einstellungen`,
`Only you can see this post so far.` → `Bisher sehen nur Sie diesen Beitrag.`,
and the plural banner in both forms. No existing translation was overwritten
(verified by diffing the catalog before and after); the only four that dropped
out are labels this feature replaced.

An AI agent wrote this text in my name, unreviewed by me. The work behind it is
mine; I only delegated the writing.

https://claude.ai/code/session_01FuNhnN7auroaXaPMFKkQCP

### DIFF
--- a/lib/vutuv/posts/photo_license.ex
+++ b/lib/vutuv/posts/photo_license.ex
@@ -79,16 +79,16 @@ defmodule Vutuv.Posts.PhotoLicense do
   def label("arr"), do: gettext("© All rights reserved")
 
   def label("cc-by-4.0"),
-    do: gettext("CC BY 4.0 — reuse with credit")
+    do: gettext("CC BY 4.0 (reuse with credit)")
 
   def label("cc-by-sa-4.0"),
-    do: gettext("CC BY-SA 4.0 — reuse with credit, share alike")
+    do: gettext("CC BY-SA 4.0 (credit, share alike)")
 
   def label("cc-by-nc-4.0"),
-    do: gettext("CC BY-NC 4.0 — non-commercial reuse with credit")
+    do: gettext("CC BY-NC 4.0 (non-commercial, with credit)")
 
   def label("cc0-1.0"),
-    do: gettext("CC0 1.0 — no rights reserved")
+    do: gettext("CC0 1.0 (no rights reserved)")
 
   def label(_unknown), do: label(@default)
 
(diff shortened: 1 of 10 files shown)


shaedrich:
@wintermeyer Would it make sense to add a general test to fail on untranslated translation strings?

wintermeyer:
> @wintermeyer Would it make sense to add a general test to fail on untranslated translation strings?

I am aware of the problem. Kicking the can down the road for now.

Issue #14, opened 2016-04-28
Avatar for User


wintermeyer:
Add a photo to a `user`.
https://github.com/stavro/arc seems to be a good solution for that.

PR #1307, opened 2026-08-02
Ein Link auf die eigene, bestätigte Webseite trägt jetzt das Häkchen (v7.224.0)

### LINKED ISSUE #1246: Mark a link in a post that points to the author's verified site
# Motivation
The Fediverse has a concept called [author attribution](https://docs.joinmastodon.org/user/profile/#attribution). We could do something similar with a user's verified links ([v7.91.0](https://github.com/wintermeyer/vutuv/commit/f92a0b502259d846fc296e40b595494176596052)):

Image

Not sure if it makes sense to show the label from the link entry here 🤔

wintermeyer:
**TL;DR** A link in a post that points at a webpage the post's author has *proved* is theirs now wears the same small emerald ✓ the profile's Links card shows — scoped so it never claims more than the proof states.

**Where:** `Vutuv.Profiles.VerifiedLinks` (new, the rule) · `VutuvWeb.Markdown.mark_verified_author_links/2` (the mark) · `Vutuv.Posts.post_preloads/0` (the data) · `VutuvWeb.AgentDocs.PostDoc` + the md/txt renderers (the siblings) · `.verified-author-link` in `assets/css/components.css`

**Now:** a member proves a profile link is really their page (`Vutuv.Profiles.LinkVerification`, v7.91.0) and it earns a mark on the profile. When they then write a post containing that same link, it looks like any other address in the text — a reader has no way to tell it is the author's own, proven site.

**Want:** mark it, and never wider than what was proved. The mark is a trust signal, so over-claiming is worse than not marking at all.

| proof | covers |
| --- | --- |
| `dns`, `well_known` | the whole host — any link on it |
| `rel_me` at the host root (`/` or no path) | the whole host |
| `rel_me` on a deeper path | **that one URL** and nothing else |

The last row is the shared-hosting case, and it is the reason the tiers exist:

```
proved by rel=me:  https://example.com/~alice
                        │
   example.com/~alice   ✓  the proven page
   example.com/~bob     ·  a different person on the same host
   example.com/~alice/foo  ·  nobody proved this
   example.com/         ·  the front page was never claimed
```

Only the **post author's** verified links are ever consulted: verification carries no uniqueness constraint (two members may each prove the same shared host by their own proof), so a global lookup would hand a stranger's proof to the wrong person.

**Normalisation:** parses with `URI.parse/1`, never a string prefix. `www.` and the apex are one party (the same reading `Vutuv.Fediverse.local_host?/1` uses), `http` and `https` are one site, and a trailing slash, the query and the fragment are ignored — a reader who pastes a link carrying `?utm_source=` is naming the same page. The path keeps its case, because a case-sensitive server really does serve `/~Alice` and `/~alice` as two pages.

**N+1:** the proven links ride in on the already-preloaded author (`post_preloads/0` scopes the `urls` preload to `verified_at is not null`), so a page of cards costs one batched query rather than one per card — asserted with `Vutuv.QueryCounter` on a five-card archive page. A surface that renders a card without that preload marks nothing rather than firing queries.

**Deliberately not done:** the profile entry's own description never replaces the anchor text — those are the author's words in their own sentence. Local posts only; the outgoing ActivityPub Note, the RSS item and `PostJSON` stay unmarked, because an emerald tick is this site's vocabulary and not something to push onto another server's renderer. Links inside code spans and fenced blocks are literal text and are never marked (the pipeline turned them into escaped text, not anchors).

**Agent formats:** the permalink's `.md`/`.txt`/`.json`/`.xml` siblings carry the same fact as data — `verified_author_links` with `address`, `url`, `verified_method` and `scope` (`"host"` / `"page"`), plus one sentence in the readable formats. Covered in `agent_docs_drift_test.exs`.

**Installability:** `:verify_user_links` off means no member has a verified link, so the list is always empty and nothing marks — it falls out, no extra gate.

**Wording** — EN `Verified webpage of the author (example.com/~alice)`, DE `Verifizierte Webseite der Autorin oder des Autors (example.com/~alice)`. Icon-only, so that is the whole `title` / `aria-label`, escaped.

`mix precommit` green: compile `--warnings-as-errors` clean, `credo --strict` found no issues, format check clean, 6680 tests pass.

Closes #1246


shaedrich:
Honestly, I find it a bit weird (and creating a lot of unused space), to have a single, small checkmark on its own line with nothing else in it:

image

If you don't want any additional verification-related text alongside the checkmark, because the tooltip says it all, I would have expected the checkmark to be a prefix of the link.

image

wintermeyer:
This seems to be a browser problem. Here is the screenshot on Safari (macOS):

Bildschirmfoto 2026-08-02 um 17 21 15

wintermeyer:
@shaedrich Please do me a favor and create new issues for these kind of bugs in the future. I saw this one by chance. I'll fix it right way. I do not read PRs after they are merged. Thanks for your help!

wintermeyer:
You were right and I was not, and I am sorry for waving this away with a Safari screenshot. That screenshot wasn't a counter-argument, it was the other half of your finding: Safari doesn't break the line in front of the tick, Chrome and Firefox do. What you saw was real, and my "this seems to be a browser problem" reply amounted to telling you that your browser was the thing at fault.

The cause is that the tick is an inline `<svg>`, so it forms an atomic inline box, and UAX #14 allows a line break in front of one. Whenever the link happened to end flush with the line, the tick went to the next line by itself. Fixed in #1309, shipped as v7.225.1: the end of the link's own text now travels with the tick inside a `white-space: nowrap` box, so it can never be stranded. Only a short slice is glued, because `nowrap` also switches off `overflow-wrap`, and a whole 40-character address in one unbreakable box scrolls a phone's post column sideways.

One thing I did keep from my side: the tick stays *after* the link rather than becoming a prefix. It marks the address you have just read, and it is the same position the profile's Links card uses, so the two surfaces say the same thing the same way. Your prefix mockup was a perfectly reasonable reading of it, though, and it is what made the wasted space unmistakable.

Thank you for reporting it, and for taking the time to draw the alternative.

PR #1508, opened 2026-08-17
Mastodon-compatible client API (v7.308.0)

jpawlowski:
Adds a Mastodon-compatible client API, so existing Mastodon phone apps can sign in to vutuv and use it as an account. Draft: the REST surface is covered by tests, but push and streaming have not been exercised against a real push service or a real client yet (see "Before merging").

## What it is

A compatibility surface served on `mastodon.<PHX_HOST>` **and on the main host**. The native `/api/2.0` contract is untouched, and the Mastodon paths (`/api/v1/…`) do not collide with it. The subdomain is the canonical origin — it is what the instance document advertises, it serves no website pages, and a reverse proxy can give it its own CORS and security headers — but a member setting up a phone app types the address they know, so `vutuv.de` has to work too. A future Pixelfed-compatible surface would get its own host beside it.

**Why both hosts rather than a redirect.** A redirect was the obvious answer and does not survive a real client: the app keeps the host you typed and builds every later URL from it, and HTTP libraries drop the `Authorization` header across a host change — the login would land and every authenticated call after it would come back 401. So both hosts answer the same routes, and the OAuth metadata names endpoints on whichever host the client arrived on, so nobody is sent across mid-flow. `/oauth/*` is deliberately not mirrored (the main host already has the real consent screen and the native token/revoke endpoints), and neither is the API catch-all, which on the main host would swallow the website.

**What a member types, and what they then see.** Server address: `vutuv.de`. The handle follows Mastodon's own `LOCAL_DOMAIN`/`WEB_DOMAIN` convention — `acct` is the bare handle, `instance.uri` is `PHX_HOST` — so a client composing the two shows `@member@vutuv.de`, the identity the fediverse already knows them by.

Covered: discovery and unattended app registration, OAuth login, accounts and relationships, statuses with photos, replies and edits, the home/public/hashtag timelines, conversations, notifications, search, the bookmark/favourite/follower/block/mute lists, profile editing, reporting, media upload, Web Push and the streaming websocket. Every list paginates Mastodon's way with `Link` headers.

## Decisions worth reviewing

**No new role.** An organization identity resolves through `Organizations.publisher?/2` — the same predicate `/act_as` uses — so a client is a second channel onto powers the Redaktion already has, never a new grant. An earlier draft of this work added `social_reader`/`social_manager` roles and a migration that wrote to `organization_roles`; that was reverted, because reshaping the role model does not belong in a change whose subject is "log in from a phone". The split is still worth making and is written up under "The staff feed" in `docs/architecture/organizations.md`.

**Nothing is invented to fill a gap.** vutuv notification kinds Mastodon has no type for (tag endorsements, CV updates, moderation cases, role grants, handle changes) are left out rather than served under a made-up type a client can neither render nor act on. `write:blocks` is never granted to an organization identity: a block is between two people.

**The public timeline is the existing site feed**, the reader behind the RSS aggregate, which lists only authors who opted out of nothing. It inherits vutuv's consent rule for aggregate surfaces rather than becoming a firehose the website does not offer.

**Media upload maps the AI image scan onto Mastodon's async media shape.** A fresh upload is pending and holds the whole post owner-only; a phone client has nowhere to render that, so `/api/v2/media` answers `202` with a null url, the client polls `GET /api/v1/media/:id` (`206`, then `200`) and posts once the picture is released — rather than publishing into a post its author alone can see.

**Web Push has no new dependency.** The obvious hex package requires `httpoison ~> 1.0`, which `CLAUDE.md` bans, and has not shipped since 2021. ECDH, HKDF and AES-128-GCM are all in OTP and delivery goes through `Req`; `mix.lock` is unchanged. A push carries no content — the notification's kind and id, nothing else — so neither the push service nor a lock screen learns what was written.

## Bugs found on the way

Two pre-existing ones, both fixed here with regression tests calibrated against the un-fixed code:

- **Attaching a photo to an organization post raised.** Such a post fills `organization_id` and leaves `user_id` nil, so `attach_images!/2` compared a column with nil, which Ecto refuses outright. The composer offers the upload whether or not you are acting as a page, so this was a 500 on an ordinary path.
- **A status could render with no text at all.** `post.images || []` let an unloaded association through the `||` and into a guard that answered an empty body. Latent until the search endpoint hit it.

## Operator-facing

- `MASTODON_API_ENABLED` (default on) turns the adapter off on both hosts; the subdomain needs its own DNS record, certificate and nginx `server_name`.
- `VAPID_PUBLIC_KEY` / `VAPID_PRIVATE_KEY` / `VAPID_SUBJECT` enable Web Push. **Unset means push is simply off**, which is what an intranet installation with no reachable push service wants.
- Both documented in `docs/ADMINS.md`.

**Migrations are additions only** — three of them, no column dropped, renamed or repurposed, nothing written to existing rows, so they are N-1 safe.

**Members and organizations are opted in by default** (`mastodon_clients?` defaults to true). Nothing is reachable without an explicit OAuth consent and the flag is a kill switch rather than the gate — but whether a new remote channel should default on for existing accounts is a product call, and I would rather it were made deliberately than inherited from this PR.

## Before merging

- `mix precommit` is green: 7,978 tests, credo strict and format clean, on Elixir 1.20 / OTP 28 / PostgreSQL 17.
- **Push, streaming and the two-host arrangement have not been smoke-tested against a real client.** The tests cover the crypto format, subscription handling, config gating, socket authentication, and that both hosts answer while the website stays intact — but no real push service and no real Mastodon app were involved. Clients differ in particular on how they compose a handle, which is the one thing only a real app answers. Per `CLAUDE.md` that is exactly the case for a device smoke test before merge.
- Rebased on `main` at v7.301.0; version bumped to **7.302.0**. Re-check it before merging — `main` moves fast enough that 7.301.0 may be taken by then.


wintermeyer (commented):
Thank you very much for this, sincerely. It is a large piece of work and the care shows. The OAuth core especially: mandatory PKCE for native clients, a sentinel that cannot be smuggled in past the length check, atomic code and refresh rotation, and a revoked grant starting fresh rather than unioning back a scope somebody took away. Better than what I would have written. Thank you too for the two pre-existing bugs you fixed on the way, with tests calibrated against the un-fixed code.

I checked the branch out. `mix precommit` is green here too. Then I probed the running app, and there are three things I would want fixed before it merges.

**1. Every paginated list except the home timeline stops at 40 items.** Measured against a 70-post source:

| endpoint | pages until the client stops | rows ever seen |
|---|---|---|
| `/timelines/home` | 20, 20, 20, 10 | 70 |
| `/accounts/:id/statuses` | 20, 20, 0 | 40 |
| `/timelines/public` | 20, 20, 0 | 40 |
| `/bookmarks` | 20, 20, 0 | 40 |

`fetch_size/1` asks for `limit + 20` newest rows and `window/3` then filters `id < max_id` in memory, so once `max_id` is older than the 40th-newest row the window is empty. Same in `/favourites`, `/timelines/tag/:hashtag`, `/accounts/:id/{followers,following}`, `/mutes`, `/statuses/:id/{favourited_by,reblogged_by}`. `followers` and `mutes` are worse: `follows_page(…, %{})` is always page one. Nothing errors, so a 200-post profile simply looks like a 40-post one.

**2. `Link` headers always name the subdomain.** `Pagination.link/3` calls `api_url/1`, not the `client_url/2` you wrote for exactly this, so a request arriving on the main host is answered with `<http://mastodon.localhost:4001/…>`. Where an operator skipped the subdomain, every "load more" fails. `StreamingSocket.check_host/1` has the same miss: still `api_host()` alone while the HTTP gate uses `client_host?/1`, so a client that signed in on the main host cannot open the stream.

**3. All engagement reads zero.** `favourites_count`, `reblogs_count` and `replies_count` are 0, and `favourited`/`reblogged`/`bookmarked` false, on every vutuv status. Measured on a post the viewer had just liked and bookmarked. The heart never fills, so a client offers "like" on an already-liked post. Accounts the same, plus an empty `note`. `Posts.shown_counts/1` has the numbers.

Smaller:

- Nothing in `test/` references `StreamingSocket`, and nothing exercises `WebPush.encrypt/3`. The description says the tests cover the crypto format and socket authentication. RFC 8291 §5 has a test vector.
- A revoked token keeps its push subscription: the cascade is on deletion, revocation is a soft `revoked_at`, and `PushDispatcher` selects on `user_id` alone.
- The route block is duplicated verbatim across both host scopes. A loop over the two hosts would remove the drift.
- `visibility: "public"` on every status, so a followers-only post reads as public and the client offers boost.
- One new German string is untranslated (`organization_live/fediverse.ex:111`). No new fuzzy entries, which is the harder half.
- `docs/ADMINS.md` still places the adapter on `mastodon.<PHX_HOST>` only. The architecture doc got the main host, the operator manual did not.
- `ListController.mutes/2` calls `follow_edge/2` per row; `follow_edges/2` is batched.
- `StatusController.delete/2` hard-matches `{:ok, _}`, so a failure is a 500 rather than JSON.
- `PushDispatcher` hardcodes `preferred_locale: "de"`.

Not yours, and already on `main`: `change_details("fediverse_changed", …)` returns `%{fields:, enabled:}` while `AccountEvents` declares only `["enabled"]`, so every fediverse settings change is rejected from the activity log. I will take that one.

The draft label is honest and the shape is right. Fix those three and I am glad to merge.

wintermeyer:
@jpawlowski I forgot to look at the location and documentation of this. I guess somewhere below /settings with a link on /settings makes most sense.

wintermeyer:
One more thing, and it may matter more than my list above: a member cannot discover any of this. The architecture doc and ADMINS.md are repo-only, `/developers` and `/llms.txt` never mention it, and the settings checkbox never names the address to type. I would like a `priv/help/` page plus the host rendered beside the toggle from `Endpoint.host()`. The feature is on by default, so it ships invisible rather than broken.

wintermeyer:
@jpawlowski 

> The feature is on by default, so it ships invisible rather than broken.

I am not 100% sure if the "on by default" makes the most sense. Probably most of the vutuv users will not use it. So my gut feeling is to deactivate it by default but put an on/off switch somewhere in the /settings So that the power users can activate it for their account.

morosanmihail:
This was something that I encountered earlier this morning. Tried to set up vutuv on a couple of Mastodon clients and they both failed. 
So being able to browse that way would be useful, at least to me.

But I do agree that, without a recommended way to interact non-web-app with vutuv, this would be more power-user focused.

jpawlowski:
Thank you — that review was worth more than the code it corrected. All three blockers are fixed, the ten smaller findings too, and the discoverability point you added last turned out to be the one that changed the most.

**Paging.** The boundary now goes into the query (`Vutuv.Keyset`), and the readers grew walkable siblings beside their offset-paged ones. `fetch_size/1` survives only for the merged home feed, which genuinely pages by a cursor, and says so. `deep_pagination_test.exs` walks ten lists with `limit=1` over 25 rows and is calibrated against the un-fixed code: with the in-memory window every case fails at exactly row 22.

Two things your table did not name and that were worse than the cap. `/accounts/:id/following` and `/statuses` could **loop forever** — a page ending on a remote row sends back `max_id=remote-<uuid>`, `cast_or_nil` answers nil, no boundary means page one again. And each list is now ordered by the id a client hands back, so the saved lists are ordered by post id rather than by when the bookmark was made; ordered by one column and paged by another, a list repeats rows and skips others, so there was no third option. The website's pagers keep their own orders.

**Hosts.** `Link` uses `client_url/2`; the socket uses the gate's `client_host?/1`.

**Engagement.** `Presenter.statuses/2` reads `post_engagement_map/2` once per page. A post carrying a denial is `private`, not `public`. One deliberate gap: the three profile figures are filled on `verify_credentials` and `/accounts/:id`, but left at zero for an account inside a list — three counts per row, and no client shows them there. `note` is filled everywhere. Say the word and I will add batched count queries.

**On/off by default:** off, for members and pages alike, as you suggested. The member's switch is on `/settings/apps` rather than the Fediverse page — federating publishes a profile to other servers, this only lets an app sign in here — and pages got their own owner-only Apps tab. Both name the address to type from `Endpoint.host()` and link the new `/system/mastodon` help page (de + en, `.md` sibling), which is also in `/llms.txt` and on `/developers`.

The rest: revoked tokens lose their push subscription (deleted on revocation, plus a token join so a later revocation path cannot forget); `mutes` reads the follow edge in one query, which also fixes its stuck first page; `delete/2` answers JSON; the push payload carries the member's locale; `StreamingSocket` and `WebPush.encrypt/5` have tests, the latter against the RFC 8291 §5 vector; ADMINS.md and the architecture doc name both hosts and the new defaults.

The duplicated route block is **gone** rather than looped — the compiler pointed out that a host-less route already matches the subdomain, so with the gate plug deciding which hosts serve the adapter, one scope does what two copies did.

One trap worth recording: `gettext.extract --merge` fuzzy-filled `"Apps – %{name}"` with `"Folgt – %{name}"`. All nineteen new German strings are written by hand.

Rebased onto `main`, version **7.304.0**, `mix precommit` green (8154 tests), `mix.lock` untouched. Out of draft — still a PR, so anything else you spot we can keep improving. The `fediverse_changed` / `AccountEvents` mismatch is untouched and yours.

*An AI agent wrote this text on my behalf.*

wintermeyer:
@jpawlowski Please resolve the conflicts. Thanks!

jpawlowski:
Conflicts resolved — rebased onto current `main`, `mergeable: MERGEABLE`, `mix precommit` green (8178 tests).

**One correction to my last comment: the version is now 7.307.0, not 7.304.0.** `main` moved from 7.301.0 to 7.306.1 while this was in review, so I rebased five times and re-bumped four. One of those was the silent kind worth knowing about: we and `main` both sat on 7.304.0, which produces no conflict and no warning — the squash would simply have landed this without moving the version. **Please re-check the number against `main` immediately before you merge**; at this cadence it can go stale again in an hour.

**One new commit since you last looked**, and it comes out of your own #1524. Before it, a post whose photo was still in the AI scan deferred its follower fan-out, which happened to keep the streaming socket correct. #1524 makes that a single fan-out at creation — correct for the website, where the reader gets a placeholder tile — but Mastodon's `update` carries a *finished* status: an attachment list has no "still processing" state and a client inserts the card once and never looks again. So a streamed post would have stayed text-only on that device forever while every other surface showed the picture. The socket now says nothing until the scan settles and then sends `status.update`, the event Mastodon has for exactly this; the client swaps the card in place. Verified against your merged code — without the guard the test goes red.

Two notes on it. The author's own pending photo is deliberately *not* handed to them: `PostImageController` authorises unreleased bytes from the browser session, and an app fetches media URLs with no credentials, so it would get the proxy's fail-closed 404 and render a broken image. And most clients never meet the window at all — Mastodon's own flow is `POST /api/v2/media`, poll until the url is non-null, *then* post, which is what our `null` url means.

Your general fediverse path needed nothing: `hold_opts/2` already delays delivery while pictures are pending and `images_settled/1` releases early. I checked before assuming.

@morosanmihail — your report is the exact case this now fixes. Typing the plain host (`vutuv.de`, no subdomain, no scheme) is what a client needs, and it works; you will have to switch app access on for your account first under Settings → Apps & API, since it now ships off. `/system/mastodon` walks through it.

Still no CI run on this PR — a fork PR needs your "Approve and run workflows".

*An AI agent wrote this text on my behalf.*

jpawlowski:
@wintermeyer Caught up again: rebased onto current `main`, `mergeable: MERGEABLE`, `mix precommit` green (8186 tests).

**Now v7.308.0**, and the PR title carries it — thanks for #1529, I had left the title on v7.301.0. One note on how that rule lands here: this PR has eight commits, so `gh pr merge --squash` will take the **title**, not a commit subject. The title is therefore the authoritative place for this one, and it is now correct.

The last collision was the silent kind again: `main` had taken 7.307.0, the same number we were carrying, so `mix.exs` did not conflict at all — both sides read identically — and the version had to be moved by hand rather than by resolving anything. That is the failure mode worth watching, because nothing reports it. Today this branch walked 7.301.0 → 7.308.0 across six rebases; every conflict in them was `mix.exs` plus the three gettext files, never application code.

Given that, a suggestion rather than a request: **you may want to set the version yourself at merge time.** You hold the merge button, so any number we write is a guess with a shelf life of about an hour, and the one shape that cannot be caught by a conflict is the one where our guess happens to match what `main` already used. If you would rather we keep chasing it, that is fine too — just say so and I will re-check before each of your merge attempts.

Everything else from your review stands as described in my two comments above. Still waiting on "Approve and run workflows" for the first CI run on this branch.

*An AI agent wrote this text on my behalf.*

wintermeyer:
Thanks for this — I read it closely and pushed fixes for all six blockers, plus a push-key crash, to `review-fixes-1508` (branched off your head). The reasoning is in the commit messages; each fix has a test calibrated against the un-fixed code, `mix precommit` is green at 8196 tests, and it re-bumps to v7.310.0 since main took 7.308.1. The two worth your eye: a phone edit cleared a post's denials, and the socket streamed raw vutuv kinds.

wintermeyer:
Merged as #1540, live in v7.311.0 — your work rebased, plus the fixes from my review. It landed by that route only because I cannot push to your fork and the PR had gone conflicting, so CI never ran; your authorship is intact on the commits. Thank you, this was excellent work.

@jpawlowski @morosanmihail — would you both point a real client at it and tell me what breaks? Push, streaming and the two-host setup have never met one.

jpawlowski:
@wintermeyer Tried it with Ivory on iPhone. Two separate failures, one ours and one infrastructure.

**1. `vutuv.de` → `unsupported_grant_type`. Our bug: we do not implement `grant_type=client_credentials`.**

Ivory churns for a few seconds and fails *before* any consent screen appears, so this is not the authorization-code exchange. Probed against production, no app registered, no side effects:

```
$ curl -s https://vutuv.de/api/v1/instance -o /dev/null -w '%{http_code}\n'
200

$ curl -s -X POST https://vutuv.de/oauth/token \
    -d 'grant_type=client_credentials&client_id=probe&client_secret=probe'
{"error":"unsupported_grant_type"}

$ curl -s -X POST https://vutuv.de/oauth/token \
    -d 'grant_type=authorization_code&client_id=probe&client_secret=probe&code=x&redirect_uri=urn:ietf:wg:oauth:2.0:oob'
{"error":"invalid_client"}
```

The second one reaching `invalid_client` is the point: that branch works and validates the client, while `client_credentials` is refused before it ever gets there. `OauthController.token/2` matches `authorization_code` and `refresh_token` and answers `unsupported_grant_type` for everything else, and `oauth_metadata` advertises `grant_types_supported: ["authorization_code"]`.

Mastodon's token endpoint **does** support `client_credentials` — it hands back an app-level token carrying the app's own scopes — and a client that wants one asks for it right after `POST /api/v1/apps`, before it sends anybody to a browser. Which grant Ivory sent is the one thing I did not capture from the device, but it was neither of the two we accept and it happened at that exact moment, so `client_credentials` is all it can reasonably have been.

Worth noting what this means for the review: the whole two-host arrangement, app registration and discovery are fine on the main host — a client gets all the way to the token endpoint and dies on a grant type. So this is a small, contained gap rather than anything structural.

Happy to implement it: accept `client_credentials`, mint a token with no user and the app's registered scopes, keep it out of every user-scoped endpoint (`Access.authorize_token/2` must refuse it rather than falling back to the app owner), and add it to `grant_types_supported`. Say the word and I will open a PR.

**2. `mastodon.vutuv.de` → "received invalid response". Not code: there is no DNS record.**

```
$ dig +short mastodon.vutuv.de A      # empty
$ dig +short mastodon.vutuv.de AAAA   # empty
$ dig +short mastodon.vutuv.de CNAME  # empty
$ dig +short vutuv.de A
134.102.58.61
```

So the subdomain needs its A/AAAA record, the TLS certificate and the nginx `server_name` before it answers. Per `docs/ADMINS.md` that subdomain is optional — the main host alone is enough for clients — so this only matters if you want the canonical origin we advertise in `oauth_metadata` to actually resolve. Until it does, `instance.uri` pointing at `PHX_HOST` is what keeps clients working.

If you would rather track the grant-type gap as its own issue than in a merged PR's thread, say so and I will move it.

*An AI agent wrote this text on my behalf.*

wintermeyer:
I wasn't aware that we need a DNS entry for mastodon.vutuv.de 
I just created that.

wintermeyer:
@jpawlowski Thanks for the Ivory run — I split the grant-type gap out into #1544 with a bug label, and yes please, go ahead and implement it along the lines you sketched.

For the next one: please open a new issue rather than commenting on a merged PR. A finding in a closed thread gets lost, while an issue stays searchable and can be assigned and labelled.

DNS for `mastodon.vutuv.de` is created.

morosanmihail:
I am getting this error currently, when trying to press "Allow access" in the Authorize view.
```
authorize?scope=read%20write%20follow&response_type=code&redirect_uri=tuba://auth_code&client_id=vu…:1 Sending form data to 'https://vutuv.de/oauth/authorize' violates the following Content Security Policy directive: "form-action 'self'". The request has been blocked.
```

morosanmihail:
The above error was for attempting to login to the account for Tuba on Linux. Tusky on Android worked fine just now.

wintermeyer:
> The above error was for attempting to login to the account for Tuba on Linux. Tusky on Android worked fine just now.

Please open an issue with a timestamp for this so that we can look it up in the log file and fix it.

PR #1540, opened 2026-08-17
Mastodon-kompatible Client-API (v7.311.0)

wintermeyer:
Julian Pawlowski's Mastodon-compatible client API from #1508, rebased onto current `main`, plus the fixes from my review of it. #1508 itself could not carry them: it sits on a fork I cannot push to, and it had gone `CONFLICTING`, which means GitHub runs no CI on it at all.

**The feature** is his, unchanged in substance: discovery and unattended app registration, OAuth login with mandatory PKCE for native clients, accounts and relationships, statuses with photos, the home/public/hashtag timelines, notifications, search, the bookmark/favourite/follower/block/mute lists, media upload mapped onto the AI image scan, Web Push with no new dependency, and the streaming websocket. Served on `mastodon.<PHX_HOST>` **and** on the main host, because an app keeps the address a member typed and HTTP libraries drop `Authorization` across a host change. Read his own write-up in #1508 for the design reasoning — it is worth the time.

**What my review changed**, seven commits, each with a test calibrated against the un-fixed code:

- **An edit from a phone cleared the post's audience and tags.** `update_post/2` replaces both wholesale (`put_assoc/3`, `on_replace: :delete`), and a Mastodon client names neither — so a post its author had closed to somebody came back public and was federated that way, silently.
- **The streaming socket spoke a different vocabulary from REST.** It sent the raw vutuv kind (`like` where `/api/v1/notifications` says `favourite`) and kinds Mastodon has no type for. One table now serves all three readers.
- **Notification paging looped forever**: the ids are `like-<uuid>`, and the prefix was not stripped off an incoming `max_id`, so "load more" served page one again.
- **The kill switch missed the websocket** — `MASTODON_API_ENABLED=false` left it accepting connections.
- **`@member@www.<host>` was treated as a foreign server**, so this installation WebFingered itself.
- **A push endpoint could point into our own network** (blind SSRF); it now runs the same two-layer `Vutuv.Ssrf` guard the webhooks use, plus `redirect: false`.
- **Smaller:** an unusable push key crashed a task on every notification, `/accounts/relationships` was unbounded, the push fan-out did two queries on the request path, and search rendered posts without their pictures or reply threading.

Six new German strings and the whole German help page used *du*; the catalogue on `main` has 0 du-forms against 548 Sie-forms, so they now match the rest of the site.

## Before merging

- `mix precommit` green: **8204 tests**, credo strict and format clean.
- **Not smoke-tested against a real Mastodon client.** Push, streaming and the two-host arrangement have never been driven by a real app. Merging on the strength of the suite is a deliberate decision, taken knowing that; jpawlowski and morosanmihail are being asked to exercise it on real clients once it is live.
- Version **7.311.0**, bumped from `main` at 7.310.1. One earlier commit subject in this branch still names 7.310.0, from before the rebase — this title is the authoritative number.


jpawlowski:
@wintermeyer have you created the DNS record for `mastodon.vutuv.de`? Seems that is still missing.

wintermeyer:
> @wintermeyer have you created the DNS record for `mastodon.vutuv.de`? Seems that is still missing.

done

PR #1546, opened 2026-08-17
A Mastodon client may ask for a token of its own (v7.315.0)

### LINKED ISSUE #1544: Mastodon API: /oauth/token rejects grant_type=client_credentials
**TL;DR** Mastodon clients cannot log in: `/oauth/token` refuses `grant_type=client_credentials`, which Mastodon's token endpoint supports.

**Where:** `VutuvWeb.OauthController.token/2` (`lib/vutuv_web/controllers/oauth_controller.ex:129`) and `grant_types_supported` in `lib/vutuv_web/controllers/mastodon_api/discovery_controller.ex:88`.
**Now:** only `authorization_code` and `refresh_token` match; everything else gets 400 `unsupported_grant_type` — Ivory on iPhone dies there, before any consent screen. Registration and discovery work, so the gap is contained.
**Want:** accept `client_credentials`, mint an app-level token with no user and the app's registered scopes, have `Access.authorize_token/2` refuse it on user-scoped endpoints rather than falling back to the app owner, and advertise the grant.
**Repro:**
```
curl -s -X POST https://vutuv.de/oauth/token \
  -d 'grant_type=client_credentials&client_id=probe&client_secret=probe'
{"error":"unsupported_grant_type"}
```

Found by @jpawlowski while testing #1508 with Ivory.

jpawlowski:
Closes #1544.

**TL;DR** A Mastodon client asks for a token of its **own** right after registering itself, and we answered `unsupported_grant_type` — which ends setup before the consent screen. `grant_type=client_credentials` (RFC 6749 §4.4) is implemented, in its own table so it can never reach a member's data. One migration (a new table, no column changed), no new dependency, no config.

## Why

@jpawlowski pointed Ivory on an iPhone at `vutuv.de`: it churned for a few seconds and gave up with `unsupported_grant_type`, never showing the consent screen. Everything downstream of that — the two hosts, `POST /api/v1/apps`, discovery — was already working.

One `curl` separated the candidates, against production, registering nothing:

```
$ curl -s -X POST https://vutuv.de/oauth/token \
    -d 'grant_type=client_credentials&client_id=probe&client_secret=probe'
{"error":"unsupported_grant_type"}

$ curl -s -X POST https://vutuv.de/oauth/token \
    -d 'grant_type=authorization_code&client_id=probe&client_secret=probe&code=x&redirect_uri=urn:ietf:wg:oauth:2.0:oob'
{"error":"invalid_client"}
```

The second reaching `invalid_client` is the point: that branch works and validates the client, while `client_credentials` was refused before the client was ever looked at. Mastodon's token endpoint answers this grant and hands back an app-level token; a client asks for one before it opens a browser.

**What is not proven:** which grant Ivory actually sent. We did not capture the request from the device, so this is the only grant it can reasonably have been, not a measured fact. If a log line from that attempt still exists, it would settle it.

## Its own table, and why that is the security property

An app token belongs to an app and to nobody, so it lands in `oauth_app_tokens` (`Vutuv.ApiAuth.AppToken`) rather than beside the member tokens.

Putting it in `api_tokens` with a null `user_id` was considered and rejected. That column is NOT NULL and `ApiAuth.lookup/1` reaches the member through an **inner join**, so a userless row would have been dropped by that join in silence — the shape CLAUDE.md records five incidents of. Widening the column would also open an N-1 window in which the previous release meets a token it cannot read.

Keeping them apart buys the property outright: every member-scoped endpoint authenticates through `api_tokens`, so an app token **cannot** be accepted by one — not because a check refuses it, but because it is not in the table that path reads. Nothing has to remember a rule, and an endpoint added to that pipeline later inherits the refusal.

What such a token may do is `GET /api/v1/apps/verify_credentials`, which names the app and echoes back neither `client_id` nor secret. It is the one route here **not** behind `Plug.MastodonApiAuth`, because that plug's job is to resolve a member; it authenticates itself. The grant is offered to `protocol: "mastodon"` apps only — a native vutuv OAuth app is user-facing with mandatory PKCE and never asked for an app credential, so it keeps the `unsupported_grant_type` it gets today.

## One deliberate difference from #1544

The issue asks for `Access.authorize_token/2` to refuse an app token rather than falling back to the app owner. This does something stronger instead: the token is not in `api_tokens` at all, so `authorize_token/2` never sees one. Every member-scoped endpoint authenticates through that table, which means an app token cannot be accepted by one — not because a function refuses it, but because it is not where that path looks. The reason for taking that route rather than the requested one is the inner join in `ApiAuth.lookup/1`: a null `user_id` row would have been dropped there in silence, and widening the column would have opened an N-1 window as well.

## Tests

14 cases. Remove the controller clause and eleven go red; the three survivors are the two asserting a refusal the fall-through already gives and the one reading the discovery document.

The second half — the app token reaching neither `verify_credentials`, a timeline, a write, a follow, nor the streaming socket, and a member token not identifying an app — is deliberately **not** calibrated against un-fixed code, and the moduledoc says so: there is no patch to revert, because the separation is structural. Those tests pin a property down rather than guard a fix.

## What still cannot be tested end to end

`mastodon.vutuv.de` now has its A record (thanks) — but it is not serving yet, so the canonical origin still cannot be exercised. Measured past the local resolver's cached negative answer:

```
$ dig +short @1.1.1.1 mastodon.vutuv.de A
134.102.58.61

$ curl -sv --resolve mastodon.vutuv.de:443:134.102.58.61 https://mastodon.vutuv.de/api/v1/instance
*  subject: CN=animina.de          # wrong certificate, handshake fails

$ curl -s -o /dev/null -w '%{http_code}\n' --resolve mastodon.vutuv.de:80:134.102.58.61 \
    http://mastodon.vutuv.de/api/v1/instance
404                                 # the request never reaches the app

$ curl -s -o /dev/null -w '%{http_code}\n' https://vutuv.de/api/v1/instance
200                                 # control: the main host is fine
```

So the subdomain needs its TLS certificate and an nginx `server_name`; right now the request lands on a default vhost serving another site. Per `docs/ADMINS.md` the subdomain is optional and the main host is enough for a client, so this does not block the fix — but it does mean the two-host arrangement stays half-tested, and push and streaming have still never met a real client.

Version **7.315.0**, `mix precommit` green (8289 tests), `mix.lock` untouched, rebased on `main`. Note that `main` had taken 7.314.0 while this was being written — the same silent collision as twice before, where both sides read the identical number so nothing conflicts — so please re-check it before merging.

*An AI agent wrote this text on my behalf.*

### DIFF
--- a/docs/architecture/mastodon-api.md
+++ b/docs/architecture/mastodon-api.md
@@ -53,9 +53,10 @@ Server discovery, login and the core social workflow:
 
 - `GET /.well-known/oauth-authorization-server`
 - `GET /api/v1/instance` and `GET /api/v2/instance`
-- `POST /api/v1/apps`
+- `POST /api/v1/apps` and `GET /api/v1/apps/verify_credentials`
 - `GET /oauth/authorize` (redirects the browser to the main origin)
-- `POST /oauth/token` and `POST /oauth/revoke`
+- `POST /oauth/token` (`authorization_code`, `refresh_token`,
+  `client_credentials`) and `POST /oauth/revoke`
 - `GET /api/v1/accounts/verify_credentials`
 - `GET /api/v1/timelines/home`
 - reading, creating, editing and deleting statuses, plus personal replies
@@ -162,6 +163,37 @@ Letting a whole workforce *read* an organization's feed without being able to
 speak in its name needs a role split that this adapter deliberately did not
 make on its way past. It is designed under "The staff feed" in
 [organizations.md](organizations.md).
+
+### The app's own token (`client_credentials`)
+
+A Mastodon client asks for a token for **itself** right after
+`POST /api/v1/apps` and before it opens a browser — RFC 6749 §4.4, which
+Mastodon's token endpoint answers. Refusing that grant does not cost a feature,
+it ends setup: Ivory pointed at `vutuv.de` churned for a few seconds and gave up
+with `unsupported_grant_type`, never reaching the consent screen, while
+everything downstream of it — the two hosts, registration, discovery — was
+already working. Diagnosing it took one `curl`: `client_credentials` answered
+`unsupported_grant_type` where `authorization_code` with the same nonsense
+credentials got as far as `invalid_client`.
+
+**Such a token lives in its own table** (`oauth_app_tokens`,
+`Vutuv.ApiAuth.AppToken`) rather than beside the member tokens, and that is the
+security property rather than a filing decision. `api_tokens.user_id` is NOT
+NULL and `ApiAuth.lookup/1` reaches the member through an **inner join**, so a
+userless row there would be dropped by that join without a word — the shape
+CLAUDE.md records five separate incidents of. Widening the column would also
+open an N-1 window in which the previous release meets a token it cannot read.
+Keeping the two apart means every member-scoped endpoint authenticates through
+`api_tokens` and therefore *cannot* accept an app token; nothing has to remember
+a rule, and an endpoint added later inherits the refusal.
+
+What such a token may do is correspondingly small:
+`GET /api/v1/apps/verify_credentials`, which names the app and deliberately
+echoes back neither `client_id` nor secret. It is the one Mastodon route not
+behind `Plug.MastodonApiAuth` — it authenticates itself, because that plug's job
+is to resolve a *member*. The grant is offered to `protocol: "mastodon"` apps
+only: a native vutuv OAuth app is user-facing with mandatory PKCE and asked for
+no app-level credential, so it keeps getting `unsupported_grant_type`.
 
 ### Paging, streaming and push
 
--- a/lib/vutuv/api_auth/app_token.ex
+++ b/lib/vutuv/api_auth/app_token.ex
@@ -1,0 +1,46 @@
+defmodule Vutuv.ApiAuth.AppToken do
+  @moduledoc """
+  A bearer credential that belongs to an **app** and to no member — RFC 6749's
+  `client_credentials` grant, which Mastodon's token endpoint answers.
+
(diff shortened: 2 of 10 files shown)


wintermeyer:
Thank you, this was both the right diagnosis and the right table, and it shipped as #1555 in v7.317.0 with your commit and authorship intact. I closed this one only because your branch lives in a fork I could not force-push into, and the rebase was not optional: while this PR conflicted, GitHub could not build its merge ref, so no CI ever ran on it.

Two things you could not have known from outside. Our production log settles the question you flagged as unproven: six attempts from that phone, each seconds after its own successful `POST /api/v1/apps`, and no `GET /oauth/authorize` anywhere in the retained window. The nginx figure needed calibrating first, because it logs the compressed body length.

And I widened one thing you had deliberately narrowed: `verify_credentials` now also answers a member token. Mastodon's documentation allows either credential there, and refusing one is the same shape of incompatibility this PR set out to remove. The separation you built is untouched, because it lives between the two tables, not between the two answers.

Beside that: the `Authorization` header is read through the plug's own parser (a lowercase `bearer` got a 401), a requested `scope` is honoured or refused by name, `/oauth/revoke` really kills an app token instead of answering 200 and doing nothing, only the newest ten rows per app survive, and `last_used_at` uses the existing throttle.

wintermeyer:
Follow-up on the one thing you flagged as unproven: it is proven now, by the client that reported it.

Since the deploy, Ivory has completed the whole flow three times, and the log ordering settles which grant it sends — each login starts with a `client_credentials` token request *before* its own `GET /oauth/authorize`, then goes through consent and redeems an authorization code. Both paths, every login. Tuba and Tusky came through the same evening.

So your reading of a failure nobody had captured was right on the first try.

PR #1695, opened 2026-08-25
Mentions read short on vutuv and in full in the Fediverse

wintermeyer:
**Symptom:** a post from another server names one of us the only way it can, `@ada@vutuv.de`, and vutuv printed that whole address on its own page. The other direction is worse: a member writes `@ada`, and on the server the post federates to, `@ada` names *their* member of that name.

**What changed:** the spelling is chosen at render time, per surface, and the stored body is never touched. `VutuvWeb.Markdown` shortens by default (`mention_form: :local`); `Vutuv.Mentions.to_local_form/1` does the same where a body is flattened to text (teaser, tab title, notification quotes). `VutuvWeb.Fediverse.Docs` renders the Note with `mention_form: :address` and adds one AP `Mention` tag per named account of ours, gated on `Fediverse.federated?/1`.

**Where to look:** any feed card naming a member, and `/:slug/posts/:id` with `Accept: application/activity+json`.

**To decide:** a topic's address (`@php@tags.vutuv.de`) is left long; shortening it to `#php` would be the same argument.

### DIFF
--- a/docs/architecture/fediverse.md
+++ b/docs/architecture/fediverse.md
@@ -1113,6 +1113,34 @@ now, with the page's own gates in front of it: the page must federate, the post
 must not be held by moderation, and a page that switched federation off gets the
 same `410`/`404` refusal its actor endpoint gives.
 
+## Mentions on the way out
+
+A member writes `@ada`, and on the server this post lands on that names *their*
+@ada. So the Note spells every mention of one of our accounts out in full —
+`@ada@vutuv.de`, the only spelling that means the same person on every server —
+while the same body keeps reading `@ada` on the page here. Nothing is rewritten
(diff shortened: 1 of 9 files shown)


wintermeyer:
Superseded by #1701, which shipped this work as v7.365.0. This PR went green and then went `CONFLICTING` on the `mix.exs` version line once the later PRs merged, and a conflicting PR gets no further CI run, so it went quiet rather than red.

#1701 carries the same two seams plus one bug found in them: `to_local_form/1` shortened inside a URL, so Mastodon's path to a remote profile, `https://mastodon.social/@ada@vutuv.de`, became `.../@ada` — a different account.

PR #1710, opened 2026-08-26
A link preview shows the page, not the cookie dialog

### LINKED ISSUE #1707: Capture the page in a link preview, not its consent dialog
**TL;DR** Link-preview captures still show a cookie dialog often enough that a reader cannot see the page. `zdf.de` is the example that started this. Neighbour of #1706 (prefer the site's own Open Graph card); this one is about the captures we keep taking.

**Where:** `Vutuv.PageScreenshot.Consent` (vendored `@duckduckgo/autoconsent`) and `Vutuv.PageScreenshot.Cdp`; rendered by `VutuvWeb.PostComponents.link_screenshot_image/1` and `VutuvWeb.UI.link_thumb/1`.

**Now:** autoconsent hides and rejects the CMPs it has a rule for; everything else is captured dialog and all, and a hopeless site is blocklisted by hand on `/admin/screenshots`. Fail-open by design, so a miss is silent.

**Want:** capture the page, not the dialog. Find out why `zdf.de` slips through, then close that class: a generic overlay/scroll-lock heuristic behind the rule set, capture only once the opt-out reports done, and a signal when a shot is probably still a dialog instead of waiting for an admin to notice.

**Repro:** post `https://www.zdf.de`, or any German news site.

jpawlowski:
**Symptom** (#1707): the preview for `zdf.de` showed "Deine Datenschutzeinstellungen" instead of the page. Not a missing consent blocker — a stale rule: autoconsent looks for `#zdf-cmp-banner-sdk`, which the rebuilt dialog no longer has. A stale rule and an unknown site fail identically, both in silence.

**What changed:** `Vutuv.PageScreenshot.ConsentSweep` runs as one `Runtime.evaluate` in the top frame just before the shutter and hides what covers the page when it is a consent notice — fixed/sticky over 4% of the viewport **and** consent vocabulary **and** an accept/reject control. CMP iframes and shadow-root CMPs are matched on id/class/src; backdrop, scroll lock and blur go with it. It hides and **never clicks**: a guessed button label is how you click "Accept".

Not in here: the signal that makes a stale rule visible, today an `:info` line nobody reads — split out as #1716.

**Smoke-tested** on real captures: zdf.de and spiegel.de show the page; tagesschau.de, github.com and Wikipedia untouched.

Hot deploy, v7.379.1. Closes #1707.


jpawlowski:
Folded into #1713.

`ConsentSweep` ships there unchanged — there was no code overlap, it sits on the CDP path before the shutter. The reason for merging the two anyway: #1713 promotes the capture from a small float beside the prose to a card's image, so a cookie dialog in it stops being a blemish and becomes the preview itself. Same acceptance bar, one PR.

#1707 is now closed by #1713.

PR #1712, opened 2026-08-26
feat(carddav): an address book that reaches the phone [60m+ ?]

### LINKED ISSUE #1705: Publish the people I follow as a read-only CardDAV book
**TL;DR** A business network is, first of all, an extended address book. The people I follow live only on vutuv — no way to subscribe to them from an iPhone, macOS Contacts or Thunderbird.

**Where:** `VutuvWeb.AgentDocs.VCard`, `Vutuv.Social.Follow`, `Vutuv.ApiAuth.AppToken`

**Now:** every profile serves one `.vcf` at `/:slug.vcf`, fetched by hand. No collection.

**Want:** a **read-only** CardDAV address book per member, members only, no organization pages, fed by the existing vCard builder, authenticated by an app-specific token, not the account password. Two new follow fields, private to its owner: a **"personally known"** flag and a free-text **note**. `/settings/carddav` picks whose cards are published, off by default:

- people I follow
- people I follow who follow me back
- only people I marked "personally known"

No bulk `.vcf` download: CardDAV can withdraw a card again, a file cannot.

jpawlowski:
**Gap:** A business network is an address book, but the people you follow never reach your phone.

**Change:** A read-only CardDAV collection at `/system/carddav/`, found via `/.well-known/carddav`. `/settings/carddav` decides whose cards you collect, four levels, "off" for everyone. `/settings/privacy` decides who may carry yours — that one **beats the subscriber's**, plus a separate level for the vCard download. A follow gains two private columns: "personally known" and a note (vCard `NOTE`). Dropped contacts leave the device next sync. Access-token sign-in, revocable per device. WebDAV-Push rides on the shared `Vutuv.WebPush`, so `MASTODON_API_ENABLED=false` does not take the address book with it. Five additive migrations, N-1 safe. New `CARDDAV_ENABLED`.

**Not included:** iOS and macOS Contacts poll instead of being pushed to — Apple's APNS extension needs a certificate no third-party server can obtain.

**Verified:** On a real iPhone — photo, note and correctly typed numbers arrive, unfollowing takes them away. Screenshots and the test-rig details are in the comments.

Closes #1705.


jpawlowski:
**Details on the test framework** (the description is capped at 150 words).

## Why it exists

**Apple never sends a password in the clear.** Contacts on iOS and macOS will not answer an HTTP Basic challenge over an unencrypted connection: they ask, they are told how to sign in, and they go quiet. The device says "account verification failed"; the server log holds eleven 401s and not one `Authorization` header. No error, no prompt. A dev server on `http://` is therefore untestable with an Apple client — and without this test CardDAV would have shipped with two bugs no green test suite could have shown.

## What it is

```bash
./scripts/e2e/tunnel.sh
```

```
device ──https──▶ Cloudflare Quick Tunnel ──▶ recording proxy ──▶ Phoenix
```

**Quick Tunnels deliberately, not ngrok**: no account, no token, no secret in the repository or in a set of instructions. One binary (`brew install cloudflared`) and one command — the only kind of tool you can ask contributors to an open-source project to install. The price is a random hostname per run and no request inspector; `scripts/e2e/inspect_proxy.mjs` replaces the latter.

Three things it gets right:

- **It waits for DNS before anything touches the name**, asking a public resolver. One query too early poisons the local resolver with a negative answer that outlives the tunnel.
- **`PHX_HOST` points at the tunnel name.** Otherwise the endpoint keeps building its links as `localhost`, and you are testing a different application through the tunnel than the one you ship. `config/dev.exs` reads the same three names for this that `runtime.exs` reads in production; unset, everything stays as before.
- **The recording never writes credentials.** An `Authorization` header becomes its scheme plus the username — enough for "did the client sign in, and as whom", and nothing reusable. Node with no dependencies, runs from a fresh checkout.

Ctrl-C stops the tunnel, the proxy and the server, and nothing else.

## What it proved

A real iPhone, iOS 27 (`dataaccessd/1.0`), from the recording:

```
OPTIONS  /system/carddav/p/<owner>/          → 200
PROPFIND /system/carddav/p/<owner>/          → 207   principal
PROPFIND /system/carddav/a/<owner>/          → 207   home
PROPFIND /system/carddav/a/<owner>/contacts/ → 207   collection
REPORT   /system/carddav/a/<owner>/contacts/ → 207   BEGIN:VCARD … NOTE:…
```

The card arrived with its `UID`, its note and a phone number, the account read-only. Then an unfollow — and the next sync-collection report carried the tombstone, a bare `404` on the card's href:

```xml
<D:response>
  <D:href>…/contacts/01a03e6b-32a9-…</D:href>
  <D:status>HTTP/1.1 404 Not Found</D:status>
</D:response>
```

The contact left the phone with nobody touching anything.

## Three bugs only this found

1. **A 301 on `/.well-known/carddav` breaks discovery.** RFC 6764 suggests the code; iOS follows it, meets the 401 at the target and does not retry there. Answered in place now.
2. **iOS asks the site root first.** `PROPFIND`/`OPTIONS` on `/` return the same service document; `GET /` stays the website.
3. **The Basic realm hung off the configured host.** A client files its password per protection space — host, port, realm, scheme. Anyone reaching the site under another name (`www.`, an installation's second hostname) finds no matching entry and silently stops. The realm is a constant now, held by a test across three hostnames.

## Beyond CardDAV

Anything that insists on TLS or has to be reachable from outside — OAuth redirects, webhooks, phone clients — can be tested against a working tree with this before it lands on `main`. Written up in `docs/DEVELOPERS.md`.

*An AI agent wrote this text on my behalf. I know that is problematic.*

jpawlowski:
#1757 (issue #1747) folds the hand-copied ETS-owner boilerplate into `use Vutuv.EtsCache`. `Uploads.AvatarCache` is the one copy it cannot reach, because it lives only on this branch — and it will not pick the shell up by itself.

No merge order is forced: the only overlap between the two is the `mix.exs` version line. Whichever lands second should convert it — `use Vutuv.EtsCache, access: :public`, `lookup/2` becomes `EtsCache.fetch/2`, `start_link/1` and the `:ets.new` call go.

wintermeyer:
Keeping this open, but it needs Stefan before it needs a reviewer: a read-only CardDAV collection is a new subsystem and a new privacy surface, not a feature, and that is his call rather than mine.

What I can say on the engineering. Making the subject's own setting beat the subscriber's is the correct direction and the thing I would have checked first — whose card may leave the site is the subject's decision, and a follower's preference must not be able to widen it. `CARDDAV_ENABLED` is the right seam for an intranet installation, and proving it against a real iPhone rather than a test client is worth more than the tests.

Two mechanical things. It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — silent, not red. And 10,800 lines across 81 files is past what anyone can review honestly in one pass; at that size a rebase also costs you more than the review does, and this branch already touches `router.ex` alongside seven other open PRs.

If Stefan says yes, split it before rebasing: the two new follow columns and their settings UI are a change that stands alone and can land first; the collection, the well-known route and the token sign-in are a second; WebDAV-Push and the illustrated guide a third. Each of those reviews in a sitting. As one PR it will keep going stale faster than it can be read.

Small thing: the footer says "on my behalf" where the others say "in my name".

jpawlowski:
Thanks — the footer now reads "in my name". Two corrections on the mechanics, both measured.

**CI is not blocked by the conflict.** The PR is `MERGEABLE` right now, and runs *are* created on every push — they park at `action_required`:

    gh run list --branch carddav --json conclusion  ->  {"action_required": 28}
    run 33241166230: event=pull_request, head_repo=jpawlowski/fork.vutuv

That is GitHub's approval gate for workflows from a **fork**, not the merge ref. All 28 sit there, including three from today while the branch was mergeable. Nothing on our side clears it — it needs someone with write access here to approve the run (or the fork-contributor setting changed). Worth knowing before anyone rebases in the hope of turning CI green: it would not.

`mix precommit` — the same alias CI runs — is green locally before every push; the pre-push hook refuses otherwise.

**The conflicts are not the branch being unusual.** Thirteen rounds in four days, and since round four it is almost always one line: `mix.exs`'s version. Splitting into three PRs multiplies that by three rather than dividing it.

**On splitting itself I have no opinion to offer** — that is scope, so it goes to Stefan with the rest. Noted that it hangs on his yes.

jpawlowski:
**The two settings pages this adds.**

carddav-settings

`/settings/carddav` — whose cards I collect. Four levels, "off" for everyone, with the contact count beside each so the choice is not abstract.

carddav-own-card

`/settings/privacy` — who may carry mine. This one beats the subscriber's setting: whoever the data is about decides. The vCard download is its own level, because a subscription and a download are different acts.

Both are the German render, which is what members see.

mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Vutuv.MixProject do
   def project do
     [
       app: :vutuv,
-      version: "7.612.0",
+      version: "7.621.0",
```
wintermeyer: Resolve this hunk to main's `version: version(),` and drop the number. Since 2026-08-29 the version is derived from the commit being built (`Vutuv.BuildInfo`), and `test/vutuv/version_test.exs` fails the build on a hand-written one. Nothing to do on your side beyond taking main's line here — the branch simply predates the change.

wintermeyer (changes requested):
This is careful work — the migrations state their N-1 reasoning, `note` is `:text` with a changeset cap, the auth plug takes a scoped token instead of the password, and the controller reads `current_user` rather than the `:member` in the path, so there is no way to open somebody else's book. The docs and the 2,000 lines of tests are there too.

It needs a rebase onto `origin/main` before I can review it properly. It is `CONFLICTING`, and a conflicting PR gets **no CI run at all** — GitHub cannot build the merge ref — which is why this one shows "no checks reported" rather than a red X. So nothing here has ever been verified.

What conflicts: `mix.exs` plus the four gettext catalogs (194 hunks in the German one alone). Two notes for that pass:

- The `mix.exs` hunk is ours, not yours — see the inline comment.
- After re-merging the catalogs, `git diff priv/gettext` and grep for `", fuzzy"` — **not** `"#, fuzzy"`, which matches nothing and reports a clean file. `--merge` fuzzy-fills new msgids with the translation of an unrelated string and nothing fails the build.

That is round 1 of 3. I have not run the security or simplify gates yet; those come once CI has actually built this.

mix.exs
jpawlowski: Done — the hunk now takes main's `version: version(),` and the number is gone. My commit touched nothing else in `mix.exs`, so nothing of mine was dropped with it. `mix test test/vutuv/version_test.exs` passes (2 tests).

Thanks for naming the cause rather than just the fix: this branch predates `Vutuv.BuildInfo`, and the version line has been the single most frequent conflict here — twenty rebases, most of them that one line. It cannot recur now.

jpawlowski:
Rebased onto current `upstream/main`, and the branch is now **one commit** — the twenty-odd `Merge upstream/main into carddav` commits are gone.

Both catalog notes are handled. I resolve the four `.po`/`.pot` conflicts by keeping **both** sides' entries and letting `mix gettext.extract --merge` normalise, then measure rather than eyeball: every upstream German and Italian entry with a non-empty `msgstr` must still be present (4,046 each, 0 missing), no duplicate on `(msgctxt, msgid)`, and `grep -c ", fuzzy"` — your spelling, not `"#, fuzzy"` — is 0 in `de` and `it`. English carries ~330 permanently, since there `msgstr == msgid`.

One correction, measured: **`CONFLICTING` is not why there is no CI.** A run is created on every push and parks at `action_required` — GitHub's approval gate for workflows from a fork. The run from 21:04 today was created while the PR was `MERGEABLE` and sits there too; 28 before it did the same. Nothing on my side clears that, and a rebase does not either. It needs someone with write access here to approve the run.

`mix precommit` — the same alias CI runs — is green locally before every push; the pre-push hook refuses otherwise.

wintermeyer:
Closing this one for its size, not for the idea.

78 files and +10,778 lines is more than I can hold in my head in one sitting, and it is CONFLICTING on top of that, so reviewing it starts with untangling it.

CardDAV is welcome back as a few PRs off current `main`: schema and store first, then the endpoint, then the sync layer. I can land those one at a time.

PR #1736, opened 2026-08-26
fix(ui): make the sticky top bar actually stick [2m]

### LINKED ISSUE #1727: Make the top bar stay put while the page scrolls
**TL;DR** The top bar carries `sticky top-0`, but its containing block is 65 px tall, so it has zero travel and scrolls away like any ordinary element, on every platform.

**Where:** `lib/vutuv_web/templates/layout/app.html.heex` (the `live_render` wrapper), `#app-shell` and the header in `VutuvWeb.ShellLive`.

**Now:** `<main>` is a *sibling* of the shell's wrapper, not a descendant. `#app-shell` therefore holds only the header — the bottom bar is `fixed` and out of flow — so its box is exactly as tall as the header. Measured against the running server: document 5342 px, `#app-shell` 65 px, sticky travel **0 px**; at `scrollY 1200` the header sits at `top: −1200`.

**Want:** `display: contents` on both wrappers, so the sticky container becomes `<body>`. Verified in a browser: the header holds at `top: 0`, the fixed bottom bar is untouched.

jpawlowski:
**What changed:** `use Phoenix.LiveView, container: {:div, class: "contents"}` on `ShellLive`, plus `class="contents"` on `#app-shell`.

**Why there:** a sticky element only slides inside its own parent's box, and both wrappers between the top bar and `<body>` were exactly as tall as the bar — zero travel. Dropping their boxes, without touching the nodes, makes `<body>` the containing block. Both are needed; removing one still leaves the bar trapped. It sits on the LiveView, not at each embedding: "my wrapper must not be a box" is a property of this shell.

**Test:** `sticky_top_bar_test.exs` asserts the rendered wrappers on both embeddings — a source grep would pass on an option a refactor stopped applying. Calibrated: dropping either half turns both cases red.

**Also verified on a running server:** desktop and 375 px, socket connected, bar at `top: 0` at `scrollY 1400`.

**Left alone:** the bar keeps `bg-white/90 backdrop-blur`.

Before and after at the same scroll position: the top bar is gone, then it stays

Closes #1727.

### DIFF
--- a/lib/vutuv_web/live/shell_live.ex
+++ b/lib/vutuv_web/live/shell_live.ex
@@ -829,7 +829,12 @@ defmodule VutuvWeb.ShellLive do
   @impl true
   def render(assigns) do
     ~H"""
-    <div id="app-shell">
+    <%!-- `contents`: the second half of the sticky fix in the `app` layout
+    (issue #1727). This div and the `live_render` container around it are the
+    two boxes between the sticky top bar and `<body>`; removing only one still
+    leaves the bar trapped in a 65px-tall parent. The layout comment carries the
+    full reasoning and the measurements. --%>
+    <div id="app-shell" class="contents">
       <%!-- Drives the green "online" dot on every avatar in the page. Receives
       this viewer's online-id set from ShellLive (push_event "presence:set") and
       writes a generated stylesheet that reveals each online member's
--- a/lib/vutuv_web/live/shell_live.ex
+++ b/lib/vutuv_web/live/shell_live.ex
@@ -12,7 +12,25 @@ defmodule VutuvWeb.ShellLive do
   read marker), messages via `Vutuv.Chat.unread_conversations_count/1`
   (conversations holding unread messages).
   """
-  use Phoenix.LiveView
+  # `container:` is what makes the top bar's `sticky top-0` mean anything
+  # (issue #1727), and it is declared here rather than at the embeddings because
+  # "my wrapper must not be a box" is a property of this shell, not of who
+  # renders it.
+  #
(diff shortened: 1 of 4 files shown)


wintermeyer:
Reviewed, and I want this — the measurements in the module comment are the part that will keep it fixed. I checked that nothing styles the container box (`#app-shell` appears in no stylesheet, and the `phx-click-loading` rules all sit on inner `data-` elements), so `display: contents` is safe here.

Two things before it can land. It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — it goes silent rather than red. Rebase onto `origin/main` and re-bump from there.

And it has no test. The failure mode is invisible: the next person who puts a box back between the bar and `<body>` breaks it with nothing going red. #1738 landed the pattern for exactly this shape — a static assertion plus a calibration you can watch fail.

jpawlowski:
Both addressed.

**Test:** `test/vutuv_web/sticky_top_bar_test.exs` asserts the *rendered* wrappers rather than the source — `container:` is a `use` option whose effect only exists once Phoenix has built the wrapper, so a grep would pass on an option a later refactor stopped applying. Both wrappers and both embeddings (`@conn`, `@socket`) are covered. Calibrated by patch: without `container:` both cases go red, and so do they without `class="contents"` on `#app-shell`.

**Conflict:** resolved by merging `origin/main` in, not rebasing — the branch is public and a force-push would rewrite what you already reviewed. Re-bumped from current main; it is `MERGEABLE` again.

wintermeyer:
I don't like a sticky top bar here. Maybe I change my opinion in the future but right now I prefer the current way.

Issue #1611, opened 2026-08-18
Let a member quote a post from another network


jpawlowski:
**TL;DR** Answering a post from another network works (#1165); quoting one does not, and quoting needs the FEP-044f consent handshake in the outbound direction.

**Where:** `Vutuv.Posts`, `VutuvWeb.PostLive.RemotePostReply` (the sibling this would follow).

**Now:** the remote post card offers reply, like and reshare. There is no quote, so a member who wants to comment on a remote post has to paste its link by hand.

**Want:** a remote-quote sidecar carrying its own copies of the target URI, actor, inbox and handle like `PostRemoteReply` does, because the cached copy expires and the quote must not. Send `QuoteRequest`, publish at once, add the authorization by `Update` when the `Accept` lands; a server that never answers leaves it pending, which the card should say. Depends on #1608 and #1610.

*An AI agent wrote this text in my name. I know that is problematic.*

Issue #1609, opened 2026-08-18
Show the quoted post when a quote arrives from another network


jpawlowski:
**TL;DR** We ignore `quote` / `quoteUri` / `_misskey_quote`, so a growing share of Mastodon 4.5 posts reaches the feed with its point missing.

**Where:** `Vutuv.Fediverse.record_remote_post/2`, `Vutuv.Fediverse.RemotePost`.

**Now:** the quoted post is neither stored nor rendered. A reader sees the commentary and a bare link, or nothing at all when the author wrote no link beside the quote.

**Want:** store the quoted URI and its `quoteAuthorization`; render a card only for a self-quote, or after verifying the stamp against the quoted object's host, and fall back to the link otherwise. A quoted remote copy needs a holder against the unfollowed purge, like a reshare or a lookup.

*An AI agent wrote this text in my name. I know that is problematic.*

jpawlowski:
We are working on this in #1743.

Scope note, because it is not obvious from the body: a quote of a **vutuv** post deliberately stays a link. `RemotePost.quote_card?/1` requires a verified FEP-044f stamp on a *remote* quoted object, and vutuv issues no stamps until #1608. Widening that gate to `quoted_local_post_id`, checked against our own authorization, is follow-up work once #1608 lands.

Phase 2 of four: #1608 → this → #1610 → #1611.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Shipped in 93917b43. A card is drawn only where consent is demonstrable — a self-quote, or a FEP-044f stamp fetched from the quoted object's own host and naming both sides; everything else stays the link it already was, including a quote of one of our own posts, which cannot verify until we issue stamps ourselves (#1608).

One thing beyond the issue: the resolution now resumes. It runs off the inbox in a task that a deploy can kill mid-fetch, so every finished attempt stamps `quote_checked_at` and a sweeper retries the unstamped rows, once each. Re-checking a stamp that already verified is #1796.

*An AI agent wrote this text in my name. I know that is problematic.*

Issue #1610, opened 2026-08-18
Let a member pass a post on with their own words


jpawlowski:
**TL;DR** vutuv has reshare, likes and replies but no quote, so passing a post on with a comment means a reshare plus a separate post of your own.

**Where:** `Vutuv.Posts`, `VutuvWeb.PostLive.ActionsComponent`.

**Now:** the two halves stay unconnected everywhere they are read — feed, profile, permalink, agent formats — and a reader has to find the original themselves.

**Want:** a quote sidecar beside `PostReply`, deliberately not a reply (out of the thread, out of the reply count, no thread notification), a compose page shaped like the answer pages, and the quoted post as a compact card one level deep. Quotable only when visible and not `restricted?`, since a quote republishes the way a reshare does.

*An AI agent wrote this text in my name. I know that is problematic.*

jpawlowski:
We are working on this in #1754 (draft).

jpawlowski:
Scope note on the four quote issues, since #1754 covers only part of the picture:

#1754 is the local half — a quote is a normal post plus a `post_quotes` sidecar, and it is **not** sent to other networks. Advertising a quote policy and issuing FEP-044f consent for our own posts is #1608; rendering an incoming quote from another network is #1609; quoting a post from another network is #1611, which depends on #1608 and this one.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1737, opened 2026-08-26
feat(fediverse): advertise a quote policy and answer QuoteRequest (FEP-044f) [15m ?]

### LINKED ISSUE #1608: Advertise a quote policy so clients offer a quote button
**TL;DR** Mastodon 4.4+ only offers a quote button when the object advertises `interactionPolicy.canQuote`, and only renders a quote once the quoted server issued a `QuoteAuthorization` (FEP-044f). We advertise neither.

**Where:** `VutuvWeb.Fediverse.Docs.note/2`, `VutuvWeb.FediverseController.perform/3`.

**Now:** our Note carries no `interactionPolicy`, and Mastodon maps a missing policy to "nobody". A `QuoteRequest` reaching our inbox is ignored, so even a forced quote stays pending forever and is never shown.

**Want:** advertise `canQuote` behind a per-member switch beside `fediverse_reactions?`, answer `QuoteRequest` with an `Accept` plus a dereferenceable stamp, and delete the stamp on takedown, freeze and audience restriction. `manualApproval` needs a request queue and is out of scope.

**Repro:** open any vutuv post in Ivory → no quote action.

jpawlowski:
**Symptom** Ivory offers no quote button on a vutuv post. Closes #1608.

**Why** Mastodon reads `interactionPolicy.canQuote` and maps an absent policy to "nobody", so no button appears. Forced, the quote stays pending forever: nothing answered the `QuoteRequest` our inbox got.

**Change** Every Note now advertises that policy — the public collection when quoting is allowed, the author alone when not. A `QuoteRequest` gets an `Accept` plus a `QuoteAuthorization` at `/:slug/posts/:id/quote-authorizations/:auth_id`. Deleting that row withdraws consent; `revoke_post/1` does it on delete, freeze and a narrowed audience. New switch on `/settings/fediverse`.

**Scope** Phase 1 of four: the answering side. Others may quote us, vutuv still cannot quote. Stands alone, merges alone — the comment says how #1609–#1611 fit.

**Decide** the default: on, because a quote redistributes an already-public post the way a reshare does.

### DIFF
--- a/docs/architecture/fediverse.md
+++ b/docs/architecture/fediverse.md
@@ -1339,6 +1339,100 @@ us. Asking it over the network would be vutuv WebFingering itself and could only
 route the visitor back here, so a signed-in member naming it simply gets the
 plain vutuv tag follow. That is `own_host?/1` again, in its third caller.
 
+## Letting other networks quote a post (issue #1608)
+
+Mastodon 4.5 authors quote posts, and it does so under **FEP-044f, "consent-
+respecting quote posts"** — the one part of the fediverse where showing somebody
+else's words is gated on that author's server saying yes. Two properties decide
+whether a vutuv post can be quoted at all, and until this issue we had neither,
+which is why a member could not quote a vutuv post from Ivory: the button was
+not offered, and forcing it would have produced a quote that stayed "pending"
+forever.
+
+**The policy is advertised, in both directions.** Every Note now carries
+
+```json
+"interactionPolicy": {"canQuote": {"automaticApproval": ["…#Public"], "manualApproval": []}}
+```
+
+when the author allows quoting, and `automaticApproval` naming the **author
+themselves** when they do not. Silence was not an option: Mastodon maps an
+absent policy to "nobody", so saying nothing is already a decision — and one a
+reader cannot tell apart from "this server has never heard of quoting". Naming
+the author rather than sending an empty list keeps it a true statement, since a
+self-quote is permitted regardless.
+
+The terms live in the document's `@context` (`gts:` for the interaction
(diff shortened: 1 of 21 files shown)


jpawlowski:
**Phase 1 of four**, the answering side: other servers may quote a vutuv post, vutuv still quotes nothing.

Rest: #1609 renders an incoming quote, #1610 quotes a vutuv post, #1611 quotes a remote one.

**Order.** #1609 and #1610 are independent of this and of each other; only #1611 needs #1608 plus #1610. So this merges alone — nothing here waits on the others.

wintermeyer:
Keeping this. Drawing a card only where the quoted author demonstrably agreed, and letting everything else stay a link, is the right default for a consent mechanism.

On the decision you asked for: **on**. Your reasoning holds — a quote redistributes an already-public post the way a reshare does, and we do not gate reshares. Stefan can overrule me, but do not hold the PR for it; ship with on and a switch.

One thing to write down before it lands: deleting the `QuoteAuthorization` row withdraws consent locally, but nothing tells the remote server, and a Mastodon that cached the authorization keeps showing the card. Say so in the module comment and in `docs/architecture/fediverse.md` — an operator who reads "deleting that row withdraws consent" will otherwise believe it took effect out there. Same shape as the takedown-revocation limit we already document.

Note it shares `lib/vutuv/fediverse.ex` with #1743, #1749 and #1778.

wintermeyer:
Closing this one. It is CONFLICTING and 53 commits behind `main`, so it no longer applies, and it overlaps both #1743 and the quote-post work in #1754.

FEP-044f is still wanted. Please rebase onto current `main` and open it again on its own, after #1743 rather than beside it, so the two can be read one at a time.

PR #1738, opened 2026-08-26
fix(mobile): stop iOS zooming the page in when a form field is focused [2m]

### LINKED ISSUE #1726: Tapping a form field zooms the page in on a phone, and it never zooms back
**TL;DR** Tapping any form field on an iPhone zooms the page to 1.14× and leaves it there, with the right edge cut off.

**Where:** `assets/css/components.css` — the `.editform` field rule (`font-size: 0.875rem`) and the base `select` rule (`15px`).

**Now:** Safari zooms whenever a focused field is under 16px, and never zooms back. Measured on iPhone 17 Pro / iOS 27: 1.14×, exactly 16 ÷ 14. Signup and every settings form are affected.

**Want:**
- 16px on every `input`, `select` and `textarea` — both rules
- `touch-action: manipulation` on links and buttons: removes double-tap zoom and the 300 ms delay, keeps pinch zoom
- `-webkit-tap-highlight-color: transparent` plus real `:active` states
- no `user-scalable=no`: it blocks deliberate zoom too (WCAG 1.4.4)

**Repro:** open `/` on a phone, tap "Vorname", close the keyboard → still zoomed.

jpawlowski:
**What changed:** an `@media (any-pointer: coarse)` block setting every text-taking field to 16px, plus `touch-action: manipulation` on links and buttons.

**Why in `app.css`, not `components.css`:** one source of the small type is `input_class/0`'s `text-sm`, a Tailwind **utility**. The components layer loses to utilities whatever its specificity, so the rule there is inert — measured, only the tag box moved. Unlayered, after `@theme`, is the one place that beats a utility.

**Why `any-pointer` and not `pointer`** (and not a width breakpoint): an iPad in landscape is wide and still zooms, and `pointer` reports only the *primary* device, so that same iPad answers `fine` the moment a keyboard is attached while the finger still zooms.

**Verified on an iPhone 17 Pro / iOS 27:** field focused, page stays at scale 1. The tests are calibrated — move the rule to a width breakpoint or back into `components.css` and they go red.

**Left out:** the tap-highlight bullet, which needs a replacement press state.

Closes #1726.


wintermeyer:
Merged as v7.507.2 — thank you, this was a pleasure to review.

Two things I want to call out. Putting the floor in `app.css` outside every cascade layer, and writing down *why* in the block's own comment, is the part I would have got wrong; and the tests are genuinely calibrated — I moved the rule to a width breakpoint and put a `font-size` back into `components.css`, and both went red with the right messages. That is rarer than it should be.

I pushed three small things to your branch before merging: the comment said three rules put a field under 16px at 14px, but the base `[type=…]` / `select` rules are 15px; the test's header claimed it held "outside every cascade layer" when it only checked the rule lives in `app.css`, so I added a second assertion that `app.css` opens no `@layer` at all; and the version needed re-bumping to 7.507.2 after main moved twice under it.

wintermeyer:
Das Problem ging mir schon lange auf den Nerv. Habe nur immer vergessen es mal anzugehen. Danke!

PR #1743, opened 2026-08-26
feat(fediverse): show a quote post from another network [20m ?]

### LINKED ISSUE #1609: Show the quoted post when a quote arrives from another network

jpawlowski:
**Symptom:** vutuv ignored `quote`, `quoteUri` and `_misskey_quote`, so a Mastodon 4.5 quote post reached the feed as commentary and a bare link, or commentary alone.

**Change:** the quoted URI and its `quoteAuthorization` are stored, and `resolve_quote/1` resolves the target off the inbox through the announce path's fence, shared rather than copied. A **card** is drawn only where the quoted author demonstrably agreed: a self-quote, or a FEP-044f stamp fetched from the quoted object's own host naming both sides. Anything else stays a link. The quoted copy is held against the unfollowed purge like a reshare.

**Resume:** that task dies with the slot a deploy stops, and says nothing. So every finished resolution stamps `quote_checked_at`, refusals included, and `Fediverse.QuoteResolver` retries the unstamped rows, once each (#1316). The migration's two foreign keys were not idempotent either.

**Not included:** quoting a vutuv post stays a link until we issue stamps (#1608); nothing re-checks a stamp that verified (#1796).

**Verified:** `mix precommit` green (9403), the resume calibrated against the un-fixed code, both states driven in a browser — screenshot below.

Closes #1609.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -501,10 +501,12 @@ config :vutuv, :fediverse_media_max_bytes, 8_000_000
 config :vutuv, :fediverse_media_fetch, true
 
 # Whether the background task that resolves what an incoming post QUOTES runs
-# (issue #1609). Off in tests, where it would touch the SQL sandbox from
-# outside; tests call Vutuv.Fediverse.resolve_quote/1 directly. Off also makes
-# every quote render as the plain link it was before this existed, which is the
-# right thing for an installation that must not make outbound requests at all.
+# (issue #1609), and with it the Vutuv.Fediverse.QuoteResolver that goes back
+# for the attempts a deploy or a crash cut short. Off in tests, where it would
+# touch the SQL sandbox from outside; tests call Vutuv.Fediverse.resolve_quote/1
+# and resume_quote_resolutions/1 directly. Off also makes every quote render as
+# the plain link it was before this existed, which is the right thing for an
+# installation that must not make outbound requests at all.
 config :vutuv, :fediverse_quote_resolve, true
 
 config :vutuv, :fediverse_remote_follow_limit, 30
--- a/config/config.exs
+++ b/config/config.exs
@@ -500,6 +500,13 @@ config :vutuv, :fediverse_media_max_bytes, 8_000_000
 # this switch is about the download, not about the display.
 config :vutuv, :fediverse_media_fetch, true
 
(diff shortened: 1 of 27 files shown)


jpawlowski:
The two states the consent gate produces, on a seeded tag page.

Top: no stamp and no self-quote, so consent was never established and it stays the plain "Zitierter Beitrag" link. Middle: consent verified, so the quoted author and their words render as a card. Bottom: the quoted post itself.

All accounts are synthetic (quote-shot.example); no real remote actor appears.

quote-states

wintermeyer:
Keeping this. Sharing the announce path's fence rather than copying it is the right instinct, and holding a quoted copy against `purge_unfollowed_remote_posts/0` the way a reshare is held is the detail that would have bitten later.

It is CONFLICTING, so no CI runs at all — silent, not red. Rebase onto `origin/main` and re-bump. The body also ends with the Claude Code footer instead of the authorship line your other PRs carry.

You name the real gap yourself: nothing re-checks a stamp, so a silent revocation keeps its card until the post expires. Do not solve it here, but give it a sweeper-shaped follow-up issue and say in the module that the card is only as fresh as the last fetch. If a recheck sweeper does land later, remember the clock rule this codebase paid for twice — stamp the `*_checked_at` column on the branch that decides an object *cannot* be checked, or that object holds the front of every batch forever.

Order for the quote milestone: #1737 (we answer), then this (we display), then #1754 (we quote). All three share `fediverse.ex` or `posts.ex`, so each rebase hands the next one work.

PR #1749, opened 2026-08-26
fix(fediverse): an answer says, out there, that it is an answer [3m]

### LINKED ISSUE #1739: Stop vutuv replies arriving in Mastodon as new threads
**TL;DR** A reply written on vutuv shows up in a Mastodon timeline as a new thread. Mastodon already hides replies to accounts you do not follow, but only when it knows a post is a reply — ours never says so.

**Where:** `VutuvWeb.Fediverse.Docs.reply_parent/1` and `cc/2` (`lib/vutuv_web/fediverse/docs.ex`).

**Now:** Two gaps. `inReplyTo` is written only when the answered author federates, and that opt-in is off by default, so most replies leave without it. And a reply to a *vutuv* post carries no `Mention` and no `cc` for the author it answers, while a reply to a *remote* post carries both.

**Want:** Write `inReplyTo` for every reply, and address the answered author the way the remote path already does. One decision comes first — see the comment below.

**Repro:** `curl -sH 'Accept: application/activity+json' <reply-permalink> | jq .inReplyTo` → `null`.

jpawlowski:
Symptom: a reply written on vutuv arrives in a Mastodon timeline as a new thread. `inReplyTo` was written only when the answered author federates, and that opt-in is off by default, so `filter_from_home` never saw a reply to hide. An answer to a vutuv post also carried no `Mention` and no `cc`.

Change: `put_in_reply_to/3` in `lib/vutuv_web/fediverse/docs.ex` now names the parent URL always. Where nobody serves a Note there, the reference dangles, deliberately: Mastodon then drops the answer from home timelines and leaves it on the profile, which beats publishing it as a fresh thread. The answered account is named in `cc`, `tag` and the leading handle only when its actor document resolves.

Not included: a restricted parent URL. Its UUID v7 id would tell readers who may not see the post that it exists and when.

Verified: `mix precommit` green; each half goes red with its fix reverted. Delivery is not gated, so `lib/vutuv/fediverse.ex` is untouched.

Closes #1739.

### DIFF
--- a/docs/architecture/fediverse.md
+++ b/docs/architecture/fediverse.md
@@ -1180,6 +1180,49 @@ that `live_render` their page get it inside
 A URL that already named an agent document is exempt, so `/jobs.md` keeps
 answering Markdown whatever `Accept` rode along with it.
 
+## An answer says it is an answer (issue #1739)
+
+An answer written here used to arrive in a Mastodon timeline looking like the
+start of a new conversation. `inReplyTo` was written only when the answered
+author federates, and `fediverse_followers?` defaults to off — so most answers
+left without it. Mastodon does hide an answer to somebody you do not follow
+(`filter_from_home` in `app/lib/feed_manager.rb`), but only once it knows a post
+*is* an answer, so these were pushed to every follower as fresh threads.
+
+Two halves, and they answer two different questions.
+
+**Does the answer say what it answers?** Always, now. `put_in_reply_to/3` names
+the parent's URL whether or not anything serves a Note at it, so the reference
+may **dangle** — that is the deliberate trade. What it costs: a reader cannot
+open the post being answered, and Mastodon drops such an answer from its *home*
+timelines outright (the `in_reply_to_id.nil?` branch of the same rule) rather
+than merely filtering it by who you follow, so it lives on the profile and at
+its own link. What it buys: nothing pretends to open a conversation it did not
+open. The member being answered kept out of the Fediverse, so their post is not
+findable out there either way.
+
+The one exception is not about resolvability but about privacy: a **restricted**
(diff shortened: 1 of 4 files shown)


wintermeyer:
Keeping this — a reply that arrives out there as a new thread is a real bug, and routing all five surfaces through one `servable_note?/1` is the right shape for it. The transitive point (holding one answer back makes its answers name an id that stopped resolving) is the part I would have missed.

On the trade you asked me to decide: take the dangling `inReplyTo`. Mastodon discards a reference it cannot resolve, so the cost is a thread that renders slightly flat on one remote server; the cost of the other choice is a reply that silently becomes a new thread for everyone, which is the bug we are fixing. Losing reach is the worse failure here because it is invisible to the member who wrote the answer.

Note it shares `lib/vutuv/fediverse.ex` with #1737, #1743 and #1778, so whichever of those lands first hands the rest a rebase.

jpawlowski:
Done — `inReplyTo` is now written unconditionally.

One correction to the premise, though: Mastodon does not discard a reference it cannot resolve. `inReplyTo` being present sets `reply: true`, and `filter_from_home` then drops the status outright when `in_reply_to_id` is nil — so such an answer is kept out of followers' home timelines entirely, not merely rendered flat. It survives on the profile and under its own link.

`servable_note?/1` and the five gates are gone; `lib/vutuv/fediverse.ex` is untouched now.

wintermeyer:
Closing this one because the core of it already landed: `main` carries the same `put_in_reply_to/3` fix for #1739 as #1841, which is also why this branch went CONFLICTING.

What is left over here is the `Mention` and the `cc`, and that is worth its own small PR off current `main`. Sorry we duplicated the work on it.

PR #1713, opened 2026-08-26
feat(posts): One card for every link, whatever the page publishes [60m+ ?]

jpawlowski:
**Gap:** every posted link got a Chromium screenshot — often of a cookie dialog, not the headline the page publishes about itself. The author never saw it before posting.

**Change:** one landscape card per link, below the post text. Each field takes the best source: picture `og:image` or our capture, headline `og:title` or `<title>`, teaser `og:description` or a locally written sentence, site `og:site_name` or the host. `PostScreenshot.card?/1` asks whether there are *words*, never where the picture came from. The composer shows the card while writing, with one button per link plus **No preview**. Additive migrations, hot deploy. Folds in #1710 and #1711.

Before, and the same post after:

Before: a small screenshot floated beside the post text

After: a landscape card with picture, domain, headline and teaser

A page with no Open Graph at all, a video, and the composer:

A link to example.com as a card: our own capture, EXAMPLE.COM, and the title element as the headline

A YouTube link as a card: the video thumbnail, YOUTUBE, and the video title

Composer showing the card and the per-link chooser

**Out of scope:** `RemoteHtml.decode_entities/1` and `scrub_nul/1` are private on `main`; `Vutuv.OpenGraph` is their second caller and opens them here. #1760 and #1764 read the columns added here.

**Verified:** each link kind composed, published and reloaded in a browser: rich Open Graph, none, video. Four new test files cover the field fallbacks and the card render.

Closes #1706, closes #1707, closes #1709, closes #1714.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1750, opened 2026-08-26
feat(feed): answers to people you do not follow stay out of the feed [5m]

### LINKED ISSUE #1740: Decide whether replies to strangers belong in the feed
**TL;DR** The vutuv newsfeed carries every reply the people you follow write, including answers to members you have never heard of. Mastodon filters exactly these out of its home timeline; we never have. This is the inward-facing half of #1739.

**Where:** `Vutuv.Posts.feed_post_items/3` (`lib/vutuv/posts.ex`).

**Now:** No reply filter at all. The discover rails carry one (`where: is_nil(r.id)`); the feed does not, so a reply and a fresh post compete on equal terms.

**Want:** A decision rather than a fix — see the comment. vutuv shows a reply with its parent inline or a "Replying to @handle" banner, so the confusion Mastodon filters against may not exist here.

**Repro:** Follow somebody who answers a stranger; their answer appears on `/`.

jpawlowski:
**Gap:** The feed carries every answer the people you follow write, including to accounts you never heard of. Read without the post it answers, such an answer is half a sentence about a stranger.

Hans answers Klaus, whom the reader does not follow — it lands in her feed:

Feed carrying an answer to an account the reader does not follow

Gone. Her own thread with Hans, and his own post, stay:

The same feed without that answer, the reader's own thread intact

**Change:** `stranger_reply_scope/2` in `Posts.feed_post_items/3` asks whether the reader is *in* this exchange, not whether she follows its author. Five ways to survive it. New switch on `/settings/preferences`, installation default at `/admin/preferences`, shipped off.

**Verified:** `mix precommit` green; the filter test fails against the un-fixed code.

**Overlaps:** `lib/vutuv/posts.ex` with #1713, #1743, #1754, #1760 and #1764 — no dependency.

Closes #1740.


wintermeyer:
Keeping this. Five ways to survive the filter, each one a conversation the reader is actually in, is the right shape — the rule is "am I in this exchange", not "do I follow the author". And thank you for catching the fuzzy fills by hand and asserting the German under `Accept-Language: de`; that trap has shipped confident nonsense here before.

On the default you flagged: **off**, as you have it. A member who opens the feed after the deploy and sees fewer strangers will not file a bug; the reverse — shipping it on and hoping people find the switch — leaves the complaint that started #1740 unanswered. Do mention it in the release note.

One wording thing, since a PR here is public: state the rule on its own terms rather than as "copies Mastodon's `filter_from_home`". Naming another network as the template we copy reads differently to an outside reader than it does to us, and the design stands on its own.

Note it shares `lib/vutuv/posts.ex` with #1713, #1743, #1754, #1760 and #1764.

wintermeyer:
Thanks — the craft here is fine, but the premise moved. #1846 shipped on 30 August: the feed now draws a reply's parent *and* the thread's root inside the same card, so an answer to a stranger is no longer half a sentence. I rendered your own example against current `main` and Klaus's question sits above Hans's answer in full.

The one reply that still reads as an orphan is one whose parent was deleted — and this filter deliberately lets that one through (`is_nil(r.parent_post_id)`).

#1740 stays open as the question it was. I left the branch alone.

PR #1751, opened 2026-08-26
fix(mobile): the current bottom tab is filled, not just tinted [2m]

### LINKED ISSUE #1728: Stop the active bottom tab being the palest of the five
**TL;DR** In the phone's bottom bar, the tab you are on is the *palest* of the five: active and inactive differ by a contrast ratio of 1.13:1, and the active one has less contrast against the bar than the inactive ones.

**Where:** `VutuvWeb.ShellLive` — `tab/1` and `nav_link_class/1`.

**Now:** active is `text-brand-600` plus `font-semibold`, inactive `text-slate-600`; same outline icon, no surface. On white: active 6.70:1, inactive 7.58:1, active against inactive **1.13:1** (dark mode 1.42:1). Below about 1.5:1 the eye reads one colour; peripheral vision keys on shape, not hue. Colour as the only carrier is also WCAG 1.4.1.

**Want:**
- a filled icon for the active state — the convention iOS and Android share; the pattern exists as `icon_scroll_top`
- weaken inactive to `slate-500` as well as strengthening active
- optionally a Material-style pill behind it

jpawlowski:
**Symptom:** on a phone the tab you are on was the palest of the five. Active `brand-600` beside inactive `slate-600` is 1.13:1, and the active tab had *less* contrast against the white bar (6.70) than its inactive neighbours (7.58). `font-semibold` at 10px carried the rest.

**Change:** in `VutuvWeb.ShellLive` the active tab draws a filled glyph — the convention iOS and Android share, because peripheral vision reads area, not hue. The magnifier has no interior, so it gets stroke weight instead; the Profile avatar a ring. Server-rendered, so it avoids the `icon_scroll_top` deploy trap.

**Not included:** the Material pill, which would fight this bar's `--press-on-bg: transparent`.

**Verified:** three new tests, each calibrated against the un-fixed code. Before/after screenshots in the comment below.

Closes #1728.


wintermeyer:
Keeping this, and it is the strongest argument in the batch: 1.13:1 is not a weak signal, it is no signal, and the active tab having *less* contrast against the bar than its neighbours is the actual bug. Filling the glyph is the right fix — area reads in peripheral vision, hue does not.

It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — silent, not red. Version 7.466.3 is now *behind* `main`, so rebase and re-bump.

One thing beyond this PR that you are better placed to sequence than I am. Six open PRs touch `shell_live.ex` — #1736, this one, #1752, #1755, #1756, #1761 — and five touch `assets/js/app.js`. Every rebase of one invalidates the next, so the order is worth choosing rather than discovering. My suggestion: this one and #1736 first (both small, both purely visual), then the tab-patching pair #1755/#1761 together since they are one idea on two surfaces, then the PWA pair.

jpawlowski:
Both shots are `/feed` on a phone viewport.

Before — all five glyphs are outlines, and the current Feed tab differs only by a slightly bluer label:

The phone tab bar on /feed before the change: all five glyphs are outlines and the current Feed tab differs only by a slightly bluer label

After — the current tab draws a solid glyph in brand blue while the other four stay outlined and grey:

The same bar after: the current Feed tab draws a solid newspaper glyph in brand blue while the other four stay outlined and grey

I moved these out of the body: at a 150-word budget the body should carry the argument, and a picture that needs a caption to be read is not saving any of it.

wintermeyer:
Merged — thank you. Your measurements carried it: the active tab having *less* contrast against the bar than its neighbours was the real bug, and the fill is the right carrier.

I pushed one commit instead of handing it back. The `refute` on Search could never fail (`icon_search` never fills), so its stroke weight and the Profile ring are pinned now, the selectors name the phone bar, and the dark press colours got the same guard.

Issue #1249, opened 2026-07-30
Browser notifications when vutuv is open but not the active tab


shaedrich:
It would be nice with vutuv could send [browser notifications](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API) if you have been online recently but your current tab isn't vutuv.

wintermeyer:
> It would be nice with vutuv could send [browser notifications](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API) if you have been online recently but your current tab isn't vutuv.

I don't understand this feature request. Please give me a concrete example.

shaedrich:
The example I've given looks concrete to me. What exactly would you know more in detail about?

1. Open your browser
2. Open a vutuv tab
3. Open another tab
4. Let someone trigger a notification in vutuv to you
5. Do not navigate back to vutuv (because you don't know about the vutuv in-app notification yet)
6. vutuv should trigger your browser to display a browser notification

wintermeyer:
In 9 out of 10 cases I hate this functionality because it is so loud. But I see your point. Let me think about it. It has to be done in a way that the people who want to use it benefit from it and others are not bothered with.

shaedrich:
I rely on that functionality a lot. Always interesting to see entirely different workflows at play 👍🏻 

Sure, it can be a setting. When the setting is activated, users are asked for permissions. Only if these two steps are consented to, the notifications will be triggered.

wintermeyer:
Shipped in v7.343.0. You were right that a setting plus a permission prompt is the shape — that is exactly what it is, with the switch off by default so nobody who doesn't want this is ever asked.

Two things I decided beyond your sketch. It fires while `document.hidden || !document.hasFocus()`, not just on a hidden tab, so a vutuv window sitting behind your editor counts as "not looking". And since the setting lives on your account while the permission lives in one browser, a browser that has never been asked gets one dismissible line offering to ask — otherwise you'd switch it on at your desk and wonder why your laptop stays quiet.

Each stream carries one notification tag, so a burst and several open tabs collapse into a single popup rather than stacking.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1752, opened 2026-08-26
feat(pwa): Installed on a phone, vutuv can now notify [30m ? shots after #1800]

### LINKED ISSUE #1729: Installed on a phone, vutuv cannot notify anyone: there is no service worker
**TL;DR** Installed on a phone, vutuv can never notify anyone: there is no service worker, so the Push API is unreachable and a tapped notification would open the browser rather than the app.

**Where:** a new `assets/js/sw.js` and a route serving it; `Vutuv.MastodonApi.WebPush`; the permission banner in `VutuvWeb.ShellLive`.

**Now:** what #1249 ships is the Notification API inside an open tab — it stops when the page closes. Real Web Push needs a `ServiceWorkerRegistration`, and there is none. The encryption half is already written: VAPID and `aes128gcm`, no extra dependency, keys from `secret_key_base` — but subscriptions hang off a mandatory `api_token_id`, so only third-party clients reach it.

**Want:**
- a small hand-written service worker: assets, offline page, `push`, `notificationclick`
- lift the crypto to `Vutuv.WebPush`, add `web_push_subscriptions` without the token
- honour `WEB_PUSH_ENABLED`

jpawlowski:
Closes #1729. Second half; #1800 carries the crypto lift and **lands first**.

**Where:** `/sw.js` (`VutuvWeb.ServiceWorkerController`), `Vutuv.WebPush.Dispatcher`.

What #1249 ships stops when the page closes, and with no worker the Push API was unreachable.

- `assets/js/sw.js`, at the root. **HTML is never cached**: digested `/assets/*` and one layout-free offline page, so no CSRF token or session is stored.
- `web_push_subscriptions` is unique on the endpoint, so a phone under a new login moves the row instead of waking its previous member.
- The payload names the kind and destination, never the text; the worker draws a generic line per kind and locale.
- `docs/architecture/realtime.md` now records what becomes of the previous release's worker.

The Browser notifications card on /settings/notifications: the account switch 'Show me browser notifications', and indented beneath it the new per-device switch 'Also notify me on this device when vutuv is closed'

The per-device switch, hidden by `WEB_PUSH_ENABLED=false`.

A brand-tinted bar above the vutuv top bar reading 'A new version of vutuv is ready.' with a Reload button

`updatefound`, with a changed worker really waiting (`reg.waiting` true).

**Decide:** the switch is off by default, and on iOS it is the only path to a notification at all. Findable enough?


wintermeyer:
Keeping this. Two decisions I want on the record because they are the ones that would have hurt: never caching HTML, so no CSRF token or session ends up in the service worker's store, and a payload that names the kind and destination but never the text. Both are the right call and neither is the obvious one.

One thing to be explicit about before it lands, because a service worker is the one asset that outlives a deploy by design: say in `docs/architecture/` what happens to a worker from the previous release that is still controlling an open tab, and for how long. You handle `updatefound` with a real waiting worker, which is the hard half — but the version skew is now a permanent property of the site rather than a transient one, and the next person to change `sw.js` needs to know that the old one is still out there.

At 3,500 lines it is also large enough that the crypto lifted out of the Mastodon adapter would review better as its own change, landed first.

wintermeyer:
Reviewed properly. This is good, and the parts I went looking for all hold up.

**The caching rule is genuinely safe.** `request.mode === "navigate"` goes straight to the network and never into the cache, `/assets/*` is cache-first only when `cacheAssets` is on (which `digested?/0` ties to a built cache manifest, so dev and test never cache a mutable bundle), and everything else gets no `respondWith` at all. The one document that *is* stored is `/system/offline`, and I checked the error layout it borrows: no CSRF token, no member, and its only `/assets` mention is a comment saying it must not depend on one. So the claim in the moduledoc is true rather than aspirational.

**Nothing member-specific is baked into `/sw.js`.** The prelude is version, `cacheAssets`, offline URL, icon, the generic line table and a fallback locale. That matters more than it looks: the worker is per-origin and outlives a sign-out, so a per-member value in there would leak across accounts on a shared device. It does not.

**The payload really is contentless** — `kind`, `locale`, `url`, nothing else.

**The migration is additive only**, and it cites the two traps it is avoiding rather than just avoiding them.

Two things I checked because they are the ones that are usually wrong:

- **The tests are not vacuous.** `config/test.exs` has `web_push_enabled: false`, which is exactly the setup where a feature's tests quietly assert nothing. They do not: each file flips the flag on itself, is `async: false` for it, uses the `fetch_env`-based `put_config` restore, and tests both sides of the gate.
- **The 45µs claim is honest.** I measured `public_key/0` at 119 reductions / ~30µs on the derived path. (My first measurement said 14 reductions / 0.1µs — that was the disabled short-circuit under the test config, not the derivation. Worth knowing if you ever benchmark this again.) Once per authenticated mount is fine; your comment's judgement stands.

**One thing I fixed and pushed.** Three of the new German strings were in the du form — the follow push line, the hint under the per-device switch, and the offline page. The catalog uses Sie in 587 places and du in none (the single grep hit is the Matrix example address `@du:matrix.org`), so this was a break in register, not a style preference. It matters most for the push line, which lands on a lock screen. I converted all three and re-ran `mix precommit`: green, 9,158 tests. No new fuzzy entries in any catalog, which given 2,148 lines of gettext churn was the other thing worth checking.

**On your question** — off by default is right, and I would not change it. An unrequested permission prompt is the fastest way to get permanently denied by the browser, and a switch the member found themselves is a prompt they will accept. Discoverability on iOS is a real gap, but it is a separate change: the honest fix is a one-time nudge inside the installed app, not moving this switch.

I rebased onto current `main` (#1798, #1799 and #1800 landed under it) and re-bumped. Note the real code here is about 1,040 lines — the rest is gettext — which is worth saying in the body, since the headline number is what decides whether anyone opens it.

PR #1753, opened 2026-08-26
feat(mobile): pull a list down to refresh it [5m]

### LINKED ISSUE #1730: Refresh the feed when it is pulled down
**TL;DR** Pulling the feed down does nothing. The gesture is what people reach for to ask "is that really everything?", and push over the socket does not replace it.

**Where:** a new hook in `assets/js`, plus a `handle_event` in `PostLive.Feed`, `MessageLive.Index` and `NotificationLive.Index`.

**Now:** content arrives over PubSub and only over PubSub. A reader cannot ask for fresh content by hand, and nothing on screen suggests they could.

**Want:**
- a hook engaging only at `scrollY === 0`, with rubber-band resistance, a threshold near 70 px and a spinner during the pull
- `navigator.vibrate` at the threshold (Android; iOS ignores it)
- **never `location.reload()`** — that discards the LiveView and reloads everything, heavier than doing nothing. `pushEvent("pwa:refresh")` instead, and the page reloads its own list
- `overscroll-behavior-y: contain`, so it does not fight Chrome's native pull

jpawlowski:
**Gap:** pulling a list down on a phone did nothing. That gesture is how people ask "is that really everything?", and no push channel answers that question — posts arrive over PubSub, but nothing confirms you have seen them all.

**Change:** a new hook (`assets/js/pull_to_refresh.js`) plus one `handle_event("pwa:refresh", …)` each in `PostLive.Feed`, `MessageLive.Index` and `NotificationLive.Index`. The release is a socket round trip, never `location.reload()` — a reload throws away the LiveView and re-fetches the document to refresh state the socket already owns. Opt-in per page: a LiveView that does not handle the event never renders the hook.

**Not included:** the visible refresh paths a page already has. The gesture is unreachable by keyboard and screen reader, so it is an addition, not a replacement.

**Verified:** tests cover the two states a still cannot show — a list already scrolled, and a pull abandoned mid-way — each calibrated against its removed guard. Driven end to end with synthetic touch events in Chrome. Screenshot in the comment below.

Closes #1730.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -42,6 +42,9 @@ import "./lightbox"
 // (self-contained; marks <html>, which is outside every LiveView root, so no
 // patch can drop the state. See scroll_top_tab.js).
 import "./scroll_top_tab"
+// Pull a list down to refresh it (issue #1730). Opt-in per page: the LiveView
+// that owns the list renders the hook and handles `pwa:refresh`.
+import PullToRefresh from "./pull_to_refresh"
 
 // LiveSocket drives the incremental LiveView shell (live unread badges, the
 // notifications/messages pages, presence). The CSRF token is rendered into the
@@ -698,6 +701,7 @@ const Hooks = {
   MarkdownEditor,
   TagInput,
   FeedUrl,
+  PullToRefresh,
   LocalTime: {
     mounted() {
       localizeTime(this.el)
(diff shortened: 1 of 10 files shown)


wintermeyer:
Keeping this. A socket round trip rather than `location.reload()` is the right release — a reload throws away the LiveView and re-fetches the document to refresh state the socket already owns. Opt-in per page, with a LiveView that does not handle the event simply not rendering the hook, is the correct shape for a gesture that only makes sense on some lists.

One thing worth a test rather than a screenshot: that the gesture does nothing when the list is already scrolled, and nothing when a pull is interrupted mid-way. Those are the two states a pull-to-refresh gets wrong, and neither shows up in a still.

jpawlowski:
A still cannot show a gesture, only what it puts on screen — the indicator part-way through the pull, just past the 70px threshold:

The feed with a small round spinner drawn below the top edge of the list, part-way through a downward pull

The two states this feature actually gets wrong are the ones no screenshot reaches, which is why they are tests rather than pictures: a list that is already scrolled never starts the gesture, and a pull the finger abandons half way refreshes nothing. Both are calibrated — remove the guard and exactly that test goes red.

wintermeyer:
Ich schliesse das mal. Irgendwie habe ich kein gutes Bauchgefühl dabei. Das Problem sollte überhaupt nicht existieren. Der neue Content wird ja vom Server automatisch gepusht.

PR #1754, opened 2026-08-26
feat(posts): pass a post on with your own words (quote posts) [25m ? after #1799]

### LINKED ISSUE #1610: Let a member pass a post on with their own words

jpawlowski:
**Lücke:** A post can be reshared, but never passed on with your own words. The only way to comment on someone else's post is a reply, which puts the comment inside their conversation.

**Change:** A quote is an ordinary post plus a `post_quotes` sidecar naming the post it carries (`Vutuv.Posts.create_quote/3`) — beside `PostReply`, deliberately not inside it: no thread, no reply count, no thread notification. The quotation-marks control on any action bar opens `/posts/:id/quote` (`VutuvWeb.PostLive.Quote`), the reply page's shape with the same composer; the quoted post renders as a compact card inside every post card, one level deep. Being quoted pins the quoted post's audience open and closes its edit window, because a quote reproduces the words rather than only carrying them away. What a reader is shown of a quoted post has one owner, `Posts.quoted_state/3`.

**Not included:** Sending a quote to other networks (FEP-044f) — #1608, #1609 and #1611 own that.

**Verified:** `mix precommit`, the agent-format drift test, and a browser pass over composing, the permalink, the profile and the archive, German and at 375px. The audience pin and the edit-window closure each have a calibrated test.

Builds on #1799, which carries the shared compose gate; merging it first makes this diff read as a feature. This branch widens the gate's name where its second caller arrives.

Closes #1610.


jpawlowski:
A quote on the post permalink, logged out: the quoter's words, the quoted post as a compact card one level deep, and the new quotation-marks control on the action bar.

quote-card

wintermeyer:
Keeping this — a sidecar beside `PostReply` rather than inside it is the right model, and giving `Posts.quoted_state/3` sole ownership of what a reader sees of a quoted post is exactly the lesson the organization milestone taught us about permission answers.

The rule I like most and want kept: being quoted pins the quoted post's audience open and closes its edit window. That is the honest reading of "a quote reproduces the words". Please make sure the edit-window closure has its own test — it is the kind of cross-feature rule that survives a refactor only if something goes red.

At 4,800 lines this is at the edge of what gets reviewed properly. If any of it separates cleanly — the shared visibility gate, say, which you describe as asked in one place for answering and quoting alike — landing that first as its own change would make the remainder a feature rather than a milestone.

jpawlowski:
Split done: #1799 carries the shared gate alone — one predicate, ~40 lines, no behaviour change. This branch sits on it and widens the name where the second caller arrives. Merge #1799 first.

The edit-window closure has its own test now, beside like/repost/reply in `posts_test.exs`, and it is calibrated: drop `has_quotes?` from `audience_locked?/1` and only that case goes red.

No rebase — five merges of `upstream/main` already cleared the conflict, and `origin` here is our fork.

wintermeyer:
Closing this one. It went CONFLICTING because its first commit is byte-for-byte the one that already shipped as #1799, so the branch and `main` both claim the same change, and 44 files is a lot to re-read once that is sorted out.

Quote posts are still wanted. Please branch fresh off current `main` without the merged commit. #1743 stays open, so keep the two apart.

PR #1755, opened 2026-08-26
feat(feed): switching tabs patches the content instead of reloading [5m]

### LINKED ISSUE #1731: Patch the content when switching bottom tabs, not the page
**TL;DR** Switching bottom tabs rebuilds the whole page — document, stylesheet, socket, shell — where a native app replaces only the content between the bars.

**Where:** `lib/vutuv_web/router.ex` (`live_session :default`, `get "/feed"`), the controllers wrapping `PostLive.Feed` and `UserProfileLive`, and the nav links in `VutuvWeb.ShellLive`.

**Now:** feed and profile already *are* LiveViews, but their controllers embed them via `live_render` so the `.md`/`.json`/`.vcf` siblings keep working — which puts them outside the `live_session`. Phoenix only patches within one session, so the nav uses plain `href`, as `shell_live.ex` says outright. Messages, notifications and search sit together and could patch today.

**Want:**
- let `VutuvWeb.Plug.AgentFormat` route extension requests to the controllers and HTML to real `live` routes
- bring feed and profile into the session
- switch nav links to `navigate`, remember scroll position per tab

jpawlowski:
**Symptom:** pressing a bottom tab rebuilt the whole page — document, stylesheet, socket, shell — where a native app replaces only the content between the bars. `/feed` was a controller route so it could negotiate its `.md/.txt/.json/.xml` siblings, and a non-`live` route cannot sit in a `live_session`.

**Change:** format separated from page. `VutuvWeb.Plug.AgentDocRoute` negotiates in a pipeline and either sends the document or falls through to a now-`live` `/feed` in `live_session :default`. `ShellLive` uses `navigate` only when **both** ends are in that session, and only once the `ShellPath` hook has reported it can say where a patch landed.

A full load always started at the top; a patch does not, so the hook resets it:

| Before | After |
|---|---|
| After pressing Feed from a scrolled notifications page, the feed opens 520px down: the tab bar sits mid-screen above empty canvas | The same press with the reset: the feed opens at the top, composer visible |

**Left out:** the profile keeps its controller; per-tab scroll *restore* is not done. Hot deploy, v7.469.0.

Closes #1731.


wintermeyer:
Closing this one as superseded by #1761 — no reflection on the work, it is the same idea and both are sound.

They close the same issue (#1731) with the same design: split format from page so `/feed` can be a `live` route inside `live_session :default`, with a plug sending the `.md/.txt/.json/.xml` siblings on. Yours calls it `AgentDocRoute`, #1761 calls it `AgentRoute`. #1761 came an hour later, and it also restores each tab's scroll position, which you explicitly left out. Reviewing both would be paying twice for one answer.

If there is anything here #1761 does not carry, say so and I will reopen rather than let it be lost.

jpawlowski:
Agreed, and nothing here is worth reopening for — #1761 is the better of the two.

I checked the three things I thought might be lost. All are covered, and two are done better: it announces the path server-side over a per-transport topic instead of my client hook, so there is no bundle capability to gate on; and it asks `ShellLive.live_session/1` through the endpoint rather than my hand-maintained `@live_paths` list, which cannot drift.

Leave it closed.

Issue #1732, opened 2026-08-26
Give the installed app a badge, shortcuts and a share target


jpawlowski:
**TL;DR** The installed app carries no unread count on its icon, offers no shortcuts on a long press, and is not in the phone's share sheet — three things people read as "a real app", none of which needs much code.

**Where:** `VutuvWeb.PageController.webmanifest/2`, plus a few lines in `assets/js/app.js`.

**Now:** the manifest declares name, icons, scope and display, and stops there.

**Want:**
- **`navigator.setAppBadge`** with the unread total. The number exists already — `ShellLive.push_badge/1` computes it for the tab title — this is the same value on the Home Screen icon
- **`shortcuts`**: feed, messages, search, new post
- **`share_target`**: vutuv in Android's share sheet, opening the composer with the shared link
- **`screenshots`** (`narrow` and `wide`), which earns a real install dialog on Android
- `display_override`, `lang`, `dir`, `description`, `categories`

*An AI agent wrote this text in my name. I know that is problematic.*

jpawlowski:
These are independent of each other and of the service worker, so they can land in any order or be split up. Ranked by effect per line of code.

### 1. Badge on the app icon

The highest ratio in the whole PWA set. `ShellLive` already computes the unread total for the tab title (`push_badge/1` → the `TabBadge` hook prefixes `document.title` with `(N)`). The same number goes on the Home Screen icon:

```js
if ('setAppBadge' in navigator) {
  count > 0 ? navigator.setAppBadge(count) : navigator.clearAppBadge()
}
```

Extend the existing hook rather than adding a second one — one owner for the number, as `push_badge/1` already is on the server. Supported in installed apps on Android and desktop, and on iOS for Home Screen web apps. Guard the call: it throws in some contexts, and it does nothing in a plain tab.

### 2. Manifest shortcuts

Long-press the icon → a menu. Pure manifest, no code:

```json
"shortcuts": [
  {"name": "Feed", "url": "/feed"},
  {"name": "Messages", "url": "/messages"},
  {"name": "Search", "url": "/search"},
  {"name": "New post", "url": "/posts/new"}
]
```

Labels must be translated, which is the one wrinkle: the manifest is served from a controller with no locale. Either serve it per `Accept-Language`, or accept English. Worth a decision, not a long one. (Check the composer's real path before using `/posts/new`.)

### 3. Share target

The strongest single "this is an app" effect. Declaring it puts vutuv in Android's share sheet next to the messengers:

[...]

PR #1756, opened 2026-08-26
feat(pwa): the installed app carries the unread count and a long-press menu [5m]

jpawlowski:
**Gap:** Installed on a phone, vutuv was a bookmark: no unread count on the icon, nothing on a long press, a bare name in the install dialog.

**Change:** the `TabBadge` hook already gets the unread total for the tab title; it now writes it with `navigator.setAppBadge` too. `#tab-badge` is the one shell element not gated on `@user_id` — the logged-out shell renders it and is pushed a zero, which clears the icon. A missing element could not tell "signed out" from "this page has no shell". `/site.webmanifest` gains four shortcuts plus `description`, `lang`, `dir`, `categories` and `display_override`; `VutuvWeb.Plug.Locale` runs on that action alone, so the labels are translated and the answer carries `Vary: accept-language`.

**Since review:** "Write a post" points at `/feed#compose`, the deep link the profile card and "n" already use — it focuses the box too, so the second `?compose=1` seam is gone.

**Not included:** `share_target`, `screenshots`. Refs #1732, one of three parts.


wintermeyer:
Keeping this, and the `#tab-badge` paragraph is the best thing in the batch: rendering the element for the logged-out shell and pushing it a zero, rather than inferring "signed out" from a missing element, is exactly right. The unsubscribe-link case you name is real — that badge would have gone stale on every page that renders no shell, and nobody would ever have reported it.

Answering `Vary: accept-language` on the manifest route is the other detail most people miss.

It is CONFLICTING, so no CI runs at all — silent, not red. Rebase onto `origin/main` and re-bump. Note it shares `shell_live.ex` with #1736, #1751, #1752 and #1761, and `app.js` with four of those; of that group this one and #1751 are the cheapest to land, so they are worth doing first.

jpawlowski:
Rebased onto `wintermeyer/vutuv` main and re-bumped: v7.526.0, one commit, MERGEABLE.

Two notes. `origin/main` on my fork is 142 versions stale (v7.365.4), so I rebased onto the upstream base you meant. And the commit subject carried a stale version — corrected in the same pass, which matters if you cherry-pick.

CI still reports nothing: fork PRs await your workflow approval, so that gate is yours, not a conflict.

wintermeyer:
Merged, thanks — this is a good one. The `#tab-badge` reasoning stands as you wrote it.

One change before merging: "Write a post" now points at `/feed#compose` instead of a new `?compose=1`. That deep link already exists (the profile's Beiträge card, the "n" shortcut) and it also puts the caret in the box, which the launcher entry wants most. So the controller/LiveView seam came back out. You lose the dead-render paint; the launcher gains a keyboard.

PR #1757, opened 2026-08-26
refactor(core): one Vutuv.EtsCache for five copies of the ETS-owner GenServer [5m]

### LINKED ISSUE #1747: Share one ETS cache instead of six copies of the same GenServer
**TL;DR** Six modules carry the same hand-copied ETS-owner GenServer.

**Where:** `Vutuv.RateLimiter`, `Vutuv.Social.PopularUsers`, `Vutuv.Posts.PopularPosts`, `Vutuv.Posts.TopPosters`, `Vutuv.SocialFeed.Cache`, `Vutuv.Uploads.AvatarCache`

**Now:** each spells out the same `start_link/1` (`Keyword.pop(opts, :name, __MODULE__)` → unregistered when `nil`, so tests get isolated instances), the same `init/1` with an injectable `table:`, and the same `lookup … rescue ArgumentError -> :miss` shell that makes a missing table a miss rather than a crash. Only the table options and the payload differ. Six places to touch for any change to how these tables are owned.

**Want:** a `use Vutuv.EtsCache, table: __MODULE__, access: :public | :protected` owning those three, and the six converted in one reviewable diff.

**Independent of #1712.** Five of the six are on `main` today; that branch only adds the sixth, following the pattern faithfully. This can go on its own schedule.

jpawlowski:
**Symptom:** Five GenServers each spelled out the same `start_link/1`, the same `init/1` table creation and the same `rescue ArgumentError -> :miss` read shell — five places to touch for any change to how these tables are owned.

**Change:** New `lib/vutuv/ets_cache.ex`. `use Vutuv.EtsCache, access: :protected | :public` generates `start_link/1` and the `open_table/1` that `init/1` calls; reads go through `EtsCache.fetch/2` and `lookup/2`. `created_by:` keeps "who may write" apart from "who calls `:ets.new/2`" — only `RateLimiter` needs it. Converted: `RateLimiter`, `Social.PopularUsers`, `Posts.TopPosters`, `SocialFeed.Cache`, `Live.MountHandoff`. No behaviour change; cold deploy.

**Not included:** the shared refresh loop (#1763) and the per-hit table probe (#1765, #1772) — each its own change.

**Merge order:** this one before #1772. It ports that probe unchanged, so landing it second would silently undo #1772. #1712 adds `Uploads.AvatarCache`; whoever lands second converts it.

**Verified:** `mix precommit` green; the four cache test files 29/29.

Closes #1747.

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -70,6 +70,12 @@ installing and operating vutuv in [Running your own vutuv](../ADMINS.md).
   exception is the people counter — the top bar's total of members here plus
   the Fediverse accounts following them — which shows the **exact** figure via
   `delimited_count/1` so it visibly ticks (see [realtime.md](realtime.md))
+- **In-memory caches**: every ETS-backed cache is a GenServer that owns one
+  named table and shares one shell, `use Vutuv.EtsCache` — its moduledoc is the
+  reference for the options. The rule worth knowing from here: a cold cache is
+  never an error. Reads go through `EtsCache.fetch/2` / `lookup/2`, which answer
+  a **miss** rather than raising when the table is not up yet, and every caller
+  has the direct query the cache exists to spare
 - **Ids**: all database ids are UUID v7 (`Vutuv.UUIDv7`): time-ordered, minted
   in the app, never integers or UUID v4.
 
--- a/lib/vutuv/ets_cache.ex
+++ b/lib/vutuv/ets_cache.ex
@@ -1,0 +1,139 @@
+defmodule Vutuv.EtsCache do
+  @moduledoc """
+  The shell every ETS-backed cache in this application shares: a GenServer that
+  owns one named table, and reads that answer a **miss** rather than crashing
+  when the table is not there.
+
+  `use Vutuv.EtsCache, access: :protected` generates the `start_link/1` and the
+  private `open_table/1` that the module's own `init/1` calls with the table it
+  wants. The reads are the plain functions `fetch/2` and `lookup/2` here. What
+  stays with each cache is what actually differs: its payload, its refresh
+  policy, and what it does with a miss.
+
+  ## Two questions, two options
+
+  **`access:` is the ETS access level — who may write.** `:protected` means the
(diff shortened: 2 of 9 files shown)


jpawlowski:
Smoke-tested on a dev server booted from this branch: all six converted GenServers come up under supervision, and the most-followed listing (`PopularUsers`) plus a profile (`TopPosters` rail, `MountHandoff` dead render into the socket connect) render unchanged.

No screenshot, on purpose: the change sits behind the caches, so every page is identical by construction and a picture would say nothing about it.

wintermeyer:
Keeping this one — six copies of the same table-owner shell is worth collapsing, and it is the foundation #1773 and #1772 both sit on.

First, the mechanical part: it is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — the PR goes silent, not red. Rebase onto `origin/main` and re-bump from there.

Then the one that matters. `open_table/1` still probes on every call: `hit_remaining/3` and `peek/2` go from `ensure_table()` to `table = open_table()`, which is the same `:ets.whereis/1` per request that #1772 exists to delete. So if this lands after #1772 it silently reverts it, and nothing goes red. Land this one first, then rebase #1772 on top so it removes the probe from `open_table/1` rather than from `ensure_table/0`.

jpawlowski:
Rebased onto `main` as a single commit and re-bumped (v7.527.1) — mergeable again.

On the probe: it is a 1:1 port of the old `ensure_table/0`, so this PR neither adds nor removes it. Agreed on the order — #1772 goes on top of this one.

One change since you looked: `PopularPosts` was removed by #1793, so this converts five modules, not six.

CI is also gated by fork approval here, not only by the conflict.

wintermeyer:
Closing this and #1773 together. Both are CONFLICTING, #1773 sits on top of this branch, and neither applies to current `main` as it stands, so the pair only reads in order.

Nothing is wrong with the change itself, and it is small. Please rebase onto current `main` and open it again, with #1773 folded in or clearly after it.

PR #1759, opened 2026-08-26
feat(mentions): suggest @-mentions while typing, and show them as chips [20m]

### LINKED ISSUE #1748: Suggest handles when @ is typed in the composer
**TL;DR** The composer offers no `@`-completion. You assemble a handle from memory and only find out after pressing Save whether it names anybody.

**Where:** `VutuvWeb.UI.markdown_editor/1` and `assets/js/markdown_editor.js`, used by the post composer (`VutuvWeb.PostLive.Composer`), replies, edit and messages. The grammar lives in `Vutuv.Mentions`.

**Now:** A mention is typed as plain text. Nothing suggests accounts while you type, and a wrong handle is caught only by `Vutuv.Mentions.validate_mentions_exist/2`, which refuses the save with "The handle @x does not exist."

**Want:** Typing `@` opens a picker of matching members and organizations — the composer already runs exactly this typeahead for its *hide from…* sheet via `Posts.search_users/2` — and the accepted mention is shown inline as a highlighted chip, the way every chat and social client has done for years. The stored source stays a bare `@handle`.

jpawlowski:
**Gap:** a mention was typed blind. The handle came from memory, and the first sign it named nobody was the save being refused by `Mentions.validate_mentions_exist/2` — after the post was written.

**Change:** typing `@` in the shared Markdown editor offers matching members and pages. `VutuvWeb.MentionController` answers two GETs under `/system/mentions/`: `suggest` ranks followed accounts first and never offers anyone blocked, hidden, unconfirmed or you; `check` resolves handles exactly as `VutuvWeb.Markdown` does when it writes the link, so a chip means "this will be a link". The document is unchanged — a bare `@handle` is inserted and the chip is decoration over it, so storage, renderer, agent formats and the federated Note see what they always saw.

**Not included:** completing remote `@user@host` addresses; the raw source view.

**Verified:** `mention_resolution_agreement_test.exs` fails if check and renderer drift; calibrated against a one-sided gate change. Driven in Chrome at `/feed`, light and dark, 375px and desktop.

Closes #1748.

### DIFF
--- a/assets/css/components.css
+++ b/assets/css/components.css
@@ -2536,6 +2536,25 @@ div.footnotes li:target {
 }
 .mde__mount .ProseMirror a { color: #2563eb; text-decoration: underline; }
 .mde__mount .ProseMirror del { text-decoration: line-through; }
+/* A confirmed `@mention` in the prose (issue #1748), drawn as the chip every
+   chat client draws: a brand-tinted pill around the handle. It is an inline
+   DECORATION over ordinary text — the stored Markdown is the bare `@handle` —
+   and it appears only once the server has confirmed the handle exists, so a
+   made-up one stays plain text instead of promising a link.
+
+   Deliberately not the published `.mention` treatment (a semibold link): on the
+   page a mention is something to follow, in the editor it is a receipt that the
+   name landed. `box-decoration-break` keeps both ends rounded when a long
+   handle wraps a line. */
+.mde__mount .ProseMirror .mde-mention {
+  background: #dbeafe;
+  border-radius: 0.35em;
+  box-decoration-break: clone;
+  -webkit-box-decoration-break: clone;
+  color: #1d4ed8;
+  font-weight: 600;
+  padding: 0.05em 0.25em;
+}
 
 /* Empty-document placeholder (see the placeholder decoration in the hook). */
 .mde__mount .ProseMirror .mde-placeholder::before {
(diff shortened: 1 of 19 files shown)


wintermeyer:
Keeping this, and "what did not change: the document" is the sentence that sells it. The picker inserting a bare `@handle` and the chip being decoration over that text means storage, the renderer, the agent formats and the federated Note all see exactly what they saw before — so this cannot break any of the four, and it can be removed later without a migration.

`check` using the same resolution `VutuvWeb.Markdown` uses to write the link is the other half: a chip that means "this will be a link" is only true while those two cannot drift. Please pin that with a test that fails if they do — one handle that resolves and one that does not, asserted through both paths. It is the kind of agreement that holds until someone adds a third account kind.

It is CONFLICTING, so no CI runs at all — silent, not red. Rebase onto `origin/main` and re-bump. Roughly 1,800 of its 3,200 lines are gettext, so the code here is a good deal smaller than the diff suggests — worth saying in the body, since the number is what decides whether anyone opens it.

jpawlowski:
Test added: `test/vutuv/mention_resolution_agreement_test.exs` asserts chip and link as an equivalence over four handles, so drift fails in either direction. The load-bearing case is a page nobody may see — it holds its handle and saves, but must not chip.

Calibrated: re-spelling the same rule leaves it green (same answer twice); only a real one-sided gate change turns it red.

Body now names the diff split — 1,700 of 3,244 lines are gettext, ~1,150 are code.

Not rebased: it is `MERGEABLE` again, kept current by merges from `main`.

jpawlowski:
Rebased as requested: 14 commits (12 of them merges from `main`) are now one commit on top of `main`, version re-drawn to v7.524.0. The tree is byte-identical to the reviewed state apart from that version line, so nothing changed except the history.

jpawlowski:
**Screenshots.** Moved out of the body per the conventions, which keep background and images in a comment.

The composer with the @-picker open: four suggested accounts for the half-typed handle @jo, the followed account first, an organization page last, and a line counting the mentions used

The same composer after accepting two mentions: @jonas_reiter and @jorunhaus sit in chips while the made-up @niemand stays plain text

There is no before shot: before this, the composer showed nothing at all for `@`, so a picture of it would assert more than it shows.

**Size.** Of 3,244 added lines, 1,700 are gettext and 315 tests — about 1,150 lines of code.

wintermeyer:
Merged, and thank you: this is lovely work, and the chip/renderer agreement test is the part I liked most.

Two things I did on your branch, so the force-push doesn't puzzle you. I rebased onto main (the gettext catalogues had drifted apart), and added one commit fixing the panel: it stayed open when you clicked beside the editor, and didn't follow the caret when the page scrolled.

Issue #1758, opened 2026-08-26
Stop the site and a client disagreeing on a post's link preview


jpawlowski:
**TL;DR** The rule "may this post carry a link preview at all" is written twice — once in the website's post card, once in the Mastodon adapter — and the two already differ in one state.

**Where:** `VutuvWeb.PostComponents.link_preview/1` + `remote_link_preview/2`, and `Vutuv.MastodonApi.Presenter.preview_card/2` (#1715).

**Now:** the *row* half has one owner (`PostScreenshot.card?/1`, `ready?/1`), the *owner* half does not. The website matches raw `images: []` and `RemotePost.warned?/1`; the adapter asks the same two questions against the attachments it names. A post whose only picture is still in the AI scan therefore has pictures to the website and none to a client, so one shows the card and the other does not.

**Want:** one predicate both call — `Screenshots.previewable_owner?(owner, images)` beside `candidate_urls/1`, which decides once which image list counts. The queue's own gate should read it too, so the queue and the renderers cannot drift.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Hi @jpawlowski — I went to reproduce the drift on `main` and the second copy of the rule is not there yet: `grep -rn "preview_card\|link_preview\|candidate_urls\|previewable_owner" lib test` returns no matches, `card:` appears exactly once in the adapter (hardcoded `nil` in `Presenter.base_status/1`), and the website half is still `assign_link_screenshot/3` over `Screenshots.qualifying_url/1`. `PostScreenshot.card?/1` and `candidate_urls/1` come with #1713, and `Presenter.preview_card/2` with #1760 stacked on top of it. So this is a follow-up to unmerged work rather than a break on `main` — with the adapter emitting no card at all today, there are not yet two answers to diverge.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Checked the premise against the branch and it holds: the website reads the raw `images` association, the adapter reads `Posts.released_images/1`, so a picture the scan still holds parts the two.

One note for whoever builds it. `previewable_owner?(owner, images)` still takes the list from the caller, and the caller is where the drift lives. Let the predicate read the list itself, and let it read the raw one: while the scan holds the picture, a picture is coming, so the card stays away.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1711, opened 2026-08-26
Hovering a link preview says what the page is about

jpawlowski:
**Symptom** (#1709): a link preview is a 400×264 picture of the top of a page — often a navigation bar and a hero image. Hovering it says nothing about where the link goes.

**What changed:** `Vutuv.LinkSummary` writes the sentence the tile's `title` now carries: the page is fetched, reduced to text by `Vutuv.RemoteHtml`, and a local Ollama model answers in ≤200 characters, in the page's own language. It lands in the new `post_screenshots.summary` after the row is already `ready`, so the picture never waits behind a model call. Not `og:description` — that is the publisher's blurb, and #1706 builds a card from it; this is for pages that publish nothing.

Best-effort, never retried, off by default (`SUMMARIZE_LINKS`). The answer is plain text, escaped and capped, and reaches a reader only as a `title`.

Cold deploy (one additive column), v7.378.0. **Unverified:** no Ollama was reachable here, so both halves ran against `plug:` stubs. Closes #1709.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1760, opened 2026-08-26
feat(mastodon-api): a link preview reaches a phone, not just a browser [5m after #1713]

### LINKED ISSUE #1715: Stop the Mastodon API returning null for every link preview
**TL;DR** A phone client reading a vutuv post through the Mastodon-compatible API never gets its link preview: the `card` field is hardcoded `nil`, so the same post shows a preview on the website and none in the app.

**Where:** `Vutuv.MastodonApi.Presenter`, the status payload's `card: nil`.

**Now:** the preview exists — `post_screenshots` carries the page's title, description, site name and our stored image since #1706 — and the presenter simply does not pass it on.

**Want:** build a Mastodon `PreviewCard` from that row (`type: "link"`, `url`, `title`, `description`, `provider_name`, `image`), `nil` where no preview is ready or the author dismissed it. Serve our own stored image URL, never the remote one — a client must not be pointed at the linked host either.

**Repro:** `GET /api/v1/statuses/:id` for a single-link post → `"card": null`.

jpawlowski:
**Gap:** a single-link post showed its preview card in a browser and a bare URL in a Mastodon client. `Vutuv.MastodonApi.Presenter` hardcoded `card: nil`, so `GET /api/v1/statuses/:id` never carried one.

**Change:** the presenter builds Mastodon's `PreviewCard` from the `post_screenshots` row already captured, for member and cached posts alike. The image it names is ours (`/screenshots/…`), never the linked host's, so reading a status in an app tells that host nothing. Its gates are the website's own predicates, so the two surfaces cannot drift. `Screenshots.preview_map/2` batches the read for a whole page.

**Not included:** no card while the AI scan holds the image, none behind a content warning. #1758 covers the owner gate still written twice.

**Verified:** `link_preview_card_test.exs` follows the card's image URL down to the file on disk; driven end to end against a dev server with a real Open Graph capture and a bearer token.

**Merge order:** #1713 first — it adds the columns this reads.

Closes #1715.


wintermeyer:
Keeping this. Naming our own capture rather than the linked host's picture is the right call and worth the sentence you gave it — reading a status in an app should tell that host nothing. Reusing the website's own predicates so the two surfaces cannot drift is the other half.

Same position as #1764: it is stacked on #1713 and carries its commits, so the 7,900 lines here are mostly not yours to defend and nobody can review it as it stands. Land #1713, rebase onto `main`, and this becomes a presenter change of a few hundred lines that reviews in one sitting.

And it is CONFLICTING, so no CI runs at all — silent, not red.

The link-preview stack is #1713 → this → #1764. Three rebases, but each starts from a `main` that already has the layer below it, which is cheaper than what three independently conflicting branches cost you today.

wintermeyer:
Closing this together with #1713 and #1764, which are the same feature in two other shapes.

This is the tangled one: 47 commits, 33 of them merges, and it pulled #1710 and #1711 in along the way, so it reads as a rival history of #1713 rather than a step on top of it. 48 of its 53 files are #1713's.

The ask is in #1713: a few small PRs off current `main`.

PR #1761, opened 2026-08-27
feat(mobile): switching bottom tabs patches the content instead of reloading [10m]

### LINKED ISSUE #1731: Patch the content when switching bottom tabs, not the page

jpawlowski:
**Symptom:** a bottom-tab press rebuilt the document, socket and shell. `/feed` was a controller route so it could serve `/feed.md`, and such a route cannot join a `live_session`.

**Change:** `VutuvWeb.Plug.AgentRoute` splits format from page — the route is `live` and names its old controller in `:private`, so the agent formats go there, HTML to the LiveView. As a routed LiveView `/feed` now inherits `:html_only` (#1776). `ShellLive` asks the router whether both ends share a live_session, and offers `navigate` only once the browser claims this bundle: a tab open across a deploy runs old JS with no `tab_scroll.js` to reset the scroll. The sticky shell learns the path over `ShellNav`.

**Not included:** the profile tab — its controller negotiates six agent formats including `.vcf`, which belongs in its own change.

**Verified:** browser smoke test on `/feed` — one `GET`, three LiveView mounts, one socket, scroll restored across a tab round trip. `mix precommit` green on every push.

Closes #1731.


jpawlowski:
**What the profile step needs**, since the PR body could not hold it.

`/:slug` has to become a `live` route, and it is the last route in the router by design — every route above wins by definition order. So the whole `live_session :default` block moves behind the `/:slug` sub-route scope, which reorders a lot of routes for one addition.

Then `UserProfileLive` has to answer from a socket mount what `UserResolveSlug` answers from a conn today: a slug that is an **organization handle** (which currently renders the page in place, at that URL), a **legacy handle** (301 to the current one) and a plain **404**. Over a socket the first two become a `Phoenix.LiveView.redirect/2` — a real browser navigation back to the same URL, where the dead render's plug does what it always did. One extra round trip on a rare path; the alternative is teaching the LiveView to be two pages.

The `<head>` work the controller does — `rel="me"`, the ActivityPub alternate, the RSS alternate, the robots header — moves to a pipeline plug, which is fine for the dead render but goes stale after a patch. Invisible to readers; crawlers do full loads.

Worth doing, worth its own PR.

wintermeyer:
Keeping this one and closing #1755 in its favour — same issue, same design, and this one also restores each tab's scroll position, which #1755 left out. Worth saying if I got that backwards.

Two small things. The body ends with the Claude Code footer rather than the authorship line the other PRs carry; please make it consistent. And "your call: should the profile tab follow?" — no, not in this PR. The profile controller owns agent-format negotiation for six formats including `.vcf`; move it in a change that is about that and nothing else.

This is the deepest PR in the shell cluster (#1736, #1751, #1752, #1756, this), so it wants to go after the two small visual ones rather than before them.

jpawlowski:
Body updated: the authorship line replaces the Claude Code footer, and the "your call" question is gone — the profile tab stays a full load and wants its own change.

One correction: it is not CONFLICTING. I merged `main` twelve times today (now v7.512.0) and GitHub reports MERGEABLE. CI awaits your approval of the fork-PR workflow, not a rebase. Merges only here, no force-push.

Ordering is yours.

wintermeyer:
Closing this one, and not because it is wrong: I read it, and the seam is sound.

But it lands routing, the app shell, tab navigation and scroll memory in one 17-file change, and I cannot give a PR that size the hours it would take to trust. My gut says no, and here that is enough.

Please re-submit it split: `AgentRoute` plus `/feed` as a live route first, the navigation and scroll work after.

PR #1771, opened 2026-08-27
fix(push): send legacy aesgcm so relay-backed iOS clients can decrypt the payload [5m]

### LINKED ISSUE #1698: Stop iOS clients failing to decrypt every push we send
### TL;DR

Ivory on iOS shows “Unable to decrypt notification” for every push from us.

### Where

`Vutuv.MastodonApi.WebPush`, `VutuvWeb.MastodonApi.PushController`

### Now vs. want

Now: we always send `Content-Encoding: aes128gcm`. Mastodon defaults its
per-subscription `standard` flag to false and sends legacy `aesgcm`, whose salt
and sender key sit in the `Encryption:`/`Crypto-Key:` headers — where the APNs
relays iOS clients need them. So the phone gets a body it cannot open. We also
drop `subscription[standard]`, so a client cannot opt in either.

Want: store `standard`, send `aesgcm` unless it is true.

### Steps to reproduce

1. Enable notifications for vutuv.de in Ivory on iOS
2. Get mentioned or followed
3. Push arrives as "Unable to decrypt notification. Rebooting your device
   should fix this problem."

### Screenshot, time, browser

Ivory for iOS. 2026-08-25, 09:25, 09:31 (4x), 11:52 (2x), 11:53 CEST. No
screenshot — the lock screen also shows unrelated private notifications.

jpawlowski:
**Symptom:** every push from us reaches Ivory on iOS as "Unable to decrypt notification". We always sent `Content-Encoding: aes128gcm`, which keeps the salt and the sender key inside the record. An iOS client sits behind an APNs relay, and APNs forwards a payload rather than our headers, so the relay has to lift those two values out of `Encryption:` and `Crypto-Key: dh=`. With `aes128gcm` there is nothing to lift.

**Change:** new column `mastodon_push_subscriptions.standard`, default false as on Mastodon. `Vutuv.WebPush.content_encoding/1` turns it into the encoding, `encode_body/4` builds it, and the new `encrypt_aesgcm/5` is the second ceremony. The installed app's own subscription carries no flag and gets `aes128gcm`, so #1752 is unaffected.

**Not included:** flipping the default. That needs someone who can confirm the relays no longer need the headers.

**Verified:** both encodings against their specs' own test vectors, plus an assertion that the shipped headers actually open the shipped body. Not verified on a physical device.

Closes #1698.


wintermeyer:
Keeping this, and it is the best-diagnosed bug in the batch. "APNs forwards a payload, not our headers, so the relay has to lift the salt and the sender key out of `Encryption:` and `Crypto-Key:` — and with `aes128gcm` there is nothing to lift" is a complete explanation in one sentence, and checking the new `encrypt_aesgcm/5` against the draft's own test vector is the right way to be sure. Asserting that the shipped headers actually open the shipped body is the part I would have skipped and regretted.

One thing to carry across, because it is easy to lose in a rebase: #1752 lifts the crypto out of `Vutuv.MastodonApi.WebPush` into `Vutuv.WebPush`, and #1712 touches that file too. Whichever of the three lands second must take the `standard` column and the `aesgcm` branch with it, or this fix quietly disappears into a refactor with every test still green. Worth a line in `docs/architecture/mastodon-api.md` saying why the default is false, so the next person moving this code knows it is a compatibility decision and not an oversight.

And a question rather than a finding: does the first-party PWA path in #1752 need the same fallback? My reading is no — Safari's own push service takes `aes128gcm`, and the Ivory case is specifically a relay rewriting the payload — but you have the device, and it is cheaper to answer now than after both have landed.

jpawlowski:
Docs line added in ee1767bb: it names `content_encoding/1` and the `aesgcm` branch of `encode_body/4`, and says why moving them goes wrong silently.

Already MERGEABLE — I merged `main` in repeatedly rather than rebasing, so no force-push.

Your reading on the PWA path matches mine: Apple's push service takes `aes128gcm`, and the choice is per subscription, so a PWA sets the flag and needs no second mechanism. I have not tested that on a device.

jpawlowski:
Tried the simulator (iPhone 17 Pro, iOS 26.5). Two measured results, neither of which answers your question:

Web Push is exposed **only to home-screen web apps** — in a Safari tab `PushManager` and `Notification` are absent entirely. That is worth knowing for #1752: the first-party path reaches installed users only.

And the simulator cannot create a subscription at all — `NotAllowedError` twice, with permission `granted` — so there is no endpoint to push to. The encoding question needs a physical device; my answer above stays reasoning, not measurement.

jpawlowski:
Continued in #1807 — same commit, same review state. GitHub closed this one because renaming the head branch (to `fix/1698-legacy-aesgcm-for-ios-relays`) drops the ref a cross-fork PR points at.

wintermeyer:
Merged. Two things I changed on the way in: the `version:` bump had to go entirely, because #1821 now derives the version from the commit and `version_test.exs` fails on a literal; and I corrected one sentence in the commit message, which named `PushSubscription.content_encoding/1` — the function is `Vutuv.WebPush.content_encoding/1`, at the send end.

Worth knowing for your other PRs: runs from your fork sit at `action_required` until I approve them, so `gh pr checks` reports none. This one had never run until today.

PR #1772, opened 2026-08-27
perf(core): skip the ETS table probe on every rate-limiter hit [2m]

### LINKED ISSUE #1765: Skip the ETS table probe on every rate-limiter hit
**TL;DR** The rate limiter probes for its ETS table before every hit; the table is always there.

**Where:** `Vutuv.RateLimiter.hit_remaining/3`, `peek/2`

**Now:** both call `ensure_table()` first, an `:ets.whereis/1` probe on every rate-limited request. `Vutuv.RateLimiter` is an unconditional supervised child, so after boot the table always exists — the lazy path is there only so a unit test can call `hit/3` with nothing supervised. `VutuvWeb.RateLimit.check/4` pays the probe two to three times per login/PIN request; the API and fediverse inbound gates once.

**Want:** the shape `VutuvWeb.Live.MountHandoff.stash/4` already uses — go straight at the table, `rescue ArgumentError -> open it, retry once`. Strictly one-shot, so a genuine badarg fails fast instead of looping.

**Overlaps #1757**, which rewrites these same lines and renames `ensure_table/0` to `open_table/0`. Whichever lands second rebases — a conflict, not a merge order.

jpawlowski:
**Symptom:** `hit_remaining/3` and `peek/2` called `ensure_table()` before every ETS operation — an `:ets.whereis/1` per rate-limited request, and `VutuvWeb.RateLimit.check/4` pays it two to three times per login. The limiter is an unconditional supervised child, so the table always exists after boot; the lazy path serves only an unsupervised unit test.

**Change:** Both go straight at the table and build it from the `ArgumentError` of a missing one — as every other ETS owner here does, so this was the last one probing. `hit_remaining/3` retries once, outside the `rescue`, so an unrelated badarg fails fast; `peek/2` needs none. Over 3M iterations: 141 -> 118 ns/op.

**Not included:** the clock read, 34 ns of the remaining 106 — a semantics change, its own issue.

**Verified:** `mix precommit` green. A test reads the compiled `hit_remaining/3` and permits only the clock and `count_hit/2`, calibrated red against `ensure_table()` and `open_table()`.

**Merge order:** #1757 restructures this file; merging it first is safer — this branch then wants `open_table()` in its two `rescue` arms.

**Deploy:** hot — no migrations, no config or dependency changes.

Closes #1765.

*Diesen Text hat ein KI-Agent in meinem Namen geschrieben. Ich weiß, dass das problematisch ist.*


jpawlowski:
Rebase note for whenever #1757 lands first: it renames `ensure_table/0` to `open_table/0` and drops the local helper, so the two `rescue` branches here call `open_table()` instead — no other change.

That resolves to `Vutuv.EtsCache.ensure(@table, :public)` under `created_by: :any_process`, i.e. race-safe table creation. Which is exactly what this change needs, since here it is a *reader* that may create the table.

jpawlowski:
`mix precommit` green on the pushed commit: compile `--warnings-as-errors`, `credo --strict` (16675 mods/funs, no issues), format, and 9075 tests, 0 failures.

Worth noting for anyone reading earlier red runs on this machine: they were contention, not this branch — same suite, 1213s at load 73 versus 269s at load 8.

wintermeyer:
Keeping this. The one-shot retry outside the `rescue` is the right shape, and the corrupt-counter test that proves it fails fast is the reason I trust the rest.

Two things. It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — the PR goes silent, not red. And its version line says 7.463.3, which is now *behind* `main` (7.507.2), so rebase and re-bump; a version below main is the one collision `bump_version.exs` cannot warn you about.

On the overlap with #1757: it is sharper than "whichever lands second rebases". #1757 replaces `ensure_table()` with `table = open_table()`, which probes on every call exactly as before — so merging it after this one **silently undoes** the optimisation, with every test still green. Take #1757 first, then rebase this onto it and delete the probe from `open_table/1`.

jpawlowski:
Both fixed. Merged `main` in rather than rebasing (no force-push on a PR under review); version is 7.516.3, and it is MERGEABLE again so CI can build the merge ref.

You are right about #1757 and I had it wrong — `open_table/1` is `EtsCache.ensure/2`, the same probe. So I pinned it: a test reads `hit_remaining/3`'s compiled calls and allows only the clock and `count_hit/2`. Calibrated red against both `ensure_table()` and `open_table()`.

Merge order and editing #1757 are yours to call.

jpawlowski:
Rebased after all — confirmed you would not be pushing here, so the force-push was safe. The nine merge commits the Autofix rounds had piled up are gone: two commits on current `main`, version 7.531.1, still MERGEABLE.

Title now follows #1774 — lowercase description. `[2m]` is 33 lines of real code, doubled for a long-lived process owning state.

No `after #1757`: no code dependency yet, and the new test makes that regression loud rather than silent.

wintermeyer:
Merged, but ahead of #1757 rather than after it as you suggested. Your own guard test is the reason: it pins `hit_remaining/3` to the clock and `count_hit/2`, so when #1757 rebases onto this, an `EtsCache.ensure/2` back on the hot path goes red instead of quietly handing the cost back with the suite green. That protection only exists in this order.

#1799 went in too — the three arms are identical to what the page asked before, and the test pins the block as the one deliberate difference.

PR #1773, opened 2026-08-27
refactor(core): one Vutuv.EtsCache.Snapshot for the copied refresh loop [5m after #1757]

### LINKED ISSUE #1763: Share one refresh loop across the three snapshot caches
**TL;DR** #1757 unified how these caches own their table. The refresh timer around it is still copied three times.

**Where:** `Vutuv.Social.PopularUsers`, `Vutuv.Posts.TopPosters`, `Vutuv.Posts.PopularPosts`

**Now:** each spells out `refresh/1`, an `init/1` reading `refresh_interval:` and firing a `refresh?`-gated 0ms seed, `handle_call(:refresh, …)`, `handle_info(:refresh, …)`, the `handle_info(_other, …)` catch-all, `schedule/1` and `enabled?/0`. Roughly 30 lines apiece, differing only in the config key (`:refresh_popular_users` / `:refresh_top_posters` / `:refresh_popular_posts`) and the body of `snapshot/1`.

**Want:** a `Vutuv.EtsCache.Snapshot` that itself `use`s `Vutuv.EtsCache`, declares `@callback snapshot(table)` and takes `refresh_every:` / `refresh_flag:`. A second module on purpose, not another option on `use Vutuv.EtsCache`: owning a table and refreshing it on a timer are separate decisions, and #1757 split them deliberately.

**Depends on #1757.** `Vutuv.SocialFeed.Cache` stays out — an on-demand single-flight fetch with a TTL sweep, not a periodic snapshot.

jpawlowski:
**Symptom:** `Vutuv.Social.PopularUsers` and `Vutuv.Posts.TopPosters` each spell out the same refresh shell — `refresh/1`, a `refresh?`-gated 0ms seed in `init/1`, both `:refresh` handlers, the catch-all, `schedule/1` and `enabled?/0`. Some 30 lines apiece, differing only in a config key and the body of `snapshot/1`.

**Change:** `Vutuv.EtsCache.Snapshot` owns that loop, declaring `@callback snapshot(table)` and taking `refresh_every:` / `refresh_flag:`. Each cache keeps its read path and the callback. A second `use` on top of `Vutuv.EtsCache`, not another option on it: owning a table and rebuilding it wholesale on a clock are separate decisions, and `RateLimiter`, `MountHandoff` and `SocialFeed.Cache` sweep rather than rebuild.

**Not included:** `Posts.PopularPosts` was the third such cache until #1793 removed it; that deletion is taken over here.

**Verified:** compile with `--warnings-as-errors`, `mix format --check-formatted`, ETS suites green. Every test calibrated against the broken code — dropping the seed, the reschedule, or moving the catch-all back into `__using__` each turns a specific test red.

**Merge order:** #1757 → this → #1772; reasoning in the pinned comment.

Closes #1763.

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -70,6 +70,16 @@ installing and operating vutuv in [Running your own vutuv](../ADMINS.md).
   exception is the people counter — the top bar's total of members here plus
   the Fediverse accounts following them — which shows the **exact** figure via
   `delimited_count/1` so it visibly ticks (see [realtime.md](realtime.md))
+- **In-memory caches**: every ETS-backed cache is a GenServer that owns one
+  named table and shares one shell, `Vutuv.EtsCache` (its moduledoc is the
+  reference): `use Vutuv.EtsCache, access: :protected | :public` generates the
+  `start_link/1` and the `open_table/1` the module's `init/1` calls, and the
+  reads go through `EtsCache.fetch/2` / `lookup/2`, which answer a **miss**
+  rather than raising when the table is not up yet. Its two options answer two
+  questions — `access:` is who may write, `created_by:` is who calls
+  `:ets.new/2` — and a cache whose readers may need the table before the
+  GenServer has it sets `created_by: :any_process`. A cold cache is never an
+  error: every caller has the direct query the cache exists to spare
 - **Ids**: all database ids are UUID v7 (`Vutuv.UUIDv7`): time-ordered, minted
   in the app, never integers or UUID v4.
 
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -79,7 +79,14 @@ installing and operating vutuv in [Running your own vutuv](../ADMINS.md).
   questions — `access:` is who may write, `created_by:` is who calls
   `:ets.new/2` — and a cache whose readers may need the table before the
   GenServer has it sets `created_by: :any_process`. A cold cache is never an
-  error: every caller has the direct query the cache exists to spare
+  error: every caller has the direct query the cache exists to spare. They all
+  tick; what differs is what the tick does. The ones whose contents are a pure
+  function of the database throw the whole table away and rebuild it, and those
+  add `Vutuv.EtsCache.Snapshot` on top — `refresh_every:` plus a `refresh_flag:`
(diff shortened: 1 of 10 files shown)


wintermeyer:
Keeping this, and the reasoning for a second `use` rather than another option on `EtsCache` is right — owning a table and recomputing it on a timer are separate decisions, and three of the six callers only want the first.

It is CONFLICTING, so no CI runs at all — silent, not red.

On order, since the ETS work is now three PRs deep on the same files: **#1757 → this → #1772**, and #1772 last for a reason. It deletes the per-call `:ets.whereis/1`, and #1757 re-introduces exactly that probe when it turns `ensure_table()` into `table = open_table()`. Landing #1772 anywhere but last means #1757 or this one silently reverts it, with every test still green. Three rebases, but each one starts from a `main` that already has the layer below it.

jpawlowski:
**Merge order: #1757 → #1773 → #1772.** Kept up to date here rather than repeated in the thread.

**#1757 first** — real code dependency. This PR's `Vutuv.EtsCache.Snapshot` is built on `Vutuv.EtsCache`, which exists only in #1757. Until it lands, this branch carries #1757's commit and its diff shows both.

**#1772 last** — not a dependency, an ordering hazard on a shared file, so it stays out of the title bracket. #1772 removes the per-call `:ets.whereis/1` from `hit_remaining/3` and `peek/2`. #1757 re-introduces exactly that probe: it rewrites both call sites to `table = open_table()`, and for `RateLimiter` (`created_by: :any_process`) `open_table/1` resolves to `EtsCache.ensure/2`, whose first statement is `:ets.whereis(name)`. Landing #1772 before #1757 therefore reverts it — the rebase applies cleanly, every test stays green, and the optimisation is silently gone.

Whoever lands second in `rate_limiter.ex` has to carry the other side over by hand.

Credit: @wintermeyer spotted the probe reversal; I verified it against both branches.

wintermeyer:
Closing this with #1757, the branch it sits on. Both are CONFLICTING, so neither applies to current `main` as it stands and the two only read in order.

The refactor itself is fine and small. Please rebase onto current `main` and open it again, either folded into #1757's successor or clearly after it.

PR #1775, opened 2026-08-27
feat(seo): organization pages and their posts preview as themselves [5m]

### LINKED ISSUE #1581: Generate contextual social preview cards for shared vutuv links
### TL;DR

Generate contextual Open Graph/social preview cards for vutuv links instead of always showing generic vutuv content

### What is hard today

Deep links to posts, profiles, or organizations currently produce almost identical previews when shared in apps like Microsoft Teams. For example, even a specific organization post may only show the generic text “Verified organization pages on vutuv…” and a generic vutuv image.

This gives no useful indication of what is actually behind the link.

### How it should work

Generate preview metadata from the actual page content:

- Posts: post text and attached image, if available
- Profiles: name, profile information, and avatar
- Organizations: organization name, description, and logo/image
- Other pages: the most relevant available content

Use the generic vutuv preview only as a fallback.

### What you are not asking for

No custom card editor or manually configured preview per post. The preview should be generated automatically from existing content.

jpawlowski:
**Gap:** an organization page, and every post published in its name, previewed with the directory's copy — "Verified organization pages on vutuv…" — and the brand card. The same card everywhere, so a shared link said nothing about what was behind it.

**Change:** `VutuvWeb.OpenGraph` gets the identity branch it only had for members: the page's name, its own description flattened out of its Markdown, its logo via `/organizations/:slug/avatar.jpg`. That tag is written only while the page is public — the endpoint 404s otherwise, and an `og:image` on a 404 costs some scrapers the whole card. A page's post permalink now previews like a member's; an attached image still beats the logo. `PostTeaser.permalink_title/1` takes sole ownership of the "Name · date" title from six surfaces that each built it, and `page_copy(["organizations" | _])` no longer lays the directory copy under every path below it.

**Not included:** cover images, per-page overrides.

**Verified:** `mix precommit` green; 28 Open Graph tests, one of them pinning the `page_copy` rule against paths rather than routes, calibrated red without the fix.

Closes #1581.


wintermeyer:
Keeping this, and it is the cleanest of the large batch. Catching that an `og:image` pointing at a 404 costs some scrapers the whole card — so the logo is only named while the page is public — is the kind of detail that decides whether the feature works at all. Giving `Vutuv.Posts.permalink_title/1` sole ownership of the "Name · date" title is right too; two call sites building the same string is exactly how the post path drifted before.

It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — silent, not red. Rebase onto `origin/main` and re-bump, and that is genuinely all it needs from me.

One check on the rebase: `page_copy(["organizations" | _])` laying the directory copy under every path below it is a bug fix hiding in a feature PR. Make sure it has its own test, so it cannot come back the next time someone adds a route under that prefix.

PR #1778, opened 2026-08-27
fix(fediverse): the host rule answers before the endpoint starts [2m]

### LINKED ISSUE #1777: Stop PeopleCounter crashing on every cold boot
**TL;DR** `Vutuv.PeopleCounter` crashes once on every cold boot, because it asks `VutuvWeb.Endpoint.host/0` before the endpoint is started.

**Where:** `lib/vutuv/people_counter.ex:215` → `Vutuv.Fediverse.distinct_follower_count/0` → `foreign_followers/0` → `own_hosts/1` (`lib/vutuv/fediverse.ex:909`).

**Now:** `init/1` seeds asynchronously with `schedule(:reconcile_fediverse, 0)`. The counter is the third child in `Vutuv.Application`; the endpoint is the last. So that message is handled before the endpoint writes its persistent term, and `Endpoint.host/0` raises there rather than answering:

```
[error] GenServer Vutuv.PeopleCounter terminating
** (RuntimeError) could not find persistent term for endpoint VutuvWeb.Endpoint.
    lib/vutuv/fediverse.ex:909: Vutuv.Fediverse.own_hosts/1
    lib/vutuv/people_counter.ex:215: Vutuv.PeopleCounter.handle_info/2
Last message: :reconcile_fediverse
```

The supervisor restarts it and the next tick recovers, so the damage is one error per boot plus a minute without a Fediverse tally — on every production cold deploy.

**Want:** the host from the endpoint's own config on that path, as `own_hosts/1`'s docstring already prescribes for the `people_snapshots` backfill, which runs in a migration and hit exactly this. Not a change to child order.

**Repro:** `mix phx.server` from a clean worktree; the crash is in the first lines after `Generated vutuv app`.

jpawlowski:
**Symptom:** `Vutuv.PeopleCounter` crashes once on every cold boot. It is the third child in `Vutuv.Application` and its zero-delay seed message reaches `VutuvWeb.Endpoint.host/0`; the endpoint is the last child, so its persistent term is missing and the call raises. One `[error]` per boot, plus a minute without a Fediverse tally.

**Change:** a private `Vutuv.Fediverse.site_host/0` asks the endpoint and, when it cannot answer yet, reads its `url: [host: …]` config — same value, available pre-boot. `own_hosts/1`, `tag_host/1` and `local_host?/1` go through it. The endpoint stays the fast path: `local_host?/1` runs ~11× per rendered post, and the config read costs 21× as much.

**Not included:** the child order, and the shipped `people_snapshots` migration that spells out the same read.

**Verified:** dev server from a clean worktree — crash before, no `[error]` after. The test erases the persistent term to rebuild the boot window; three of its four cases fail against the un-fixed code.

Hot deploy, v7.560.3.

Closes #1777.


wintermeyer:
Keeping this — a crash on every cold boot is worth fixing, and the calibration test that first proves the endpoint really is unavailable is what makes the other three mean anything. I checked the rescue type against the dependency: Phoenix's generated `persistent!/0` does a bare `raise "could not find persistent term…"`, so `RuntimeError` is right, and `:persistent_term.get(key, nil)` means there is no `ArgumentError` hiding underneath.

It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — the PR goes silent rather than red, which reads exactly like a broken workflow. Rebase onto `origin/main` and re-bump from there; that is all this one needs from me.

jpawlowski:
Thanks — the rescue-type check is the bit I most wanted a second pair of eyes on.

The conflict is already gone: I merged `main` in rather than rebasing, and the PR has been MERGEABLE since v7.513.1.

That did not start CI, and a rebase would not either. `gh run list` shows a CI run created for **every** push, each `completed / action_required / 0s` — the fork "Approve and run workflows" gate, not a missing merge ref. One click on your side runs it.

PR #1780, opened 2026-08-27
test(core): share one mount-session helper across the off-router LiveViews [2m]

### LINKED ISSUE #1779: Share one mount-session helper across the shell's tests
**TL;DR** The mount session for `VutuvWeb.ShellLive` is spelled out in six test files, so a change to `Vutuv.Sessions.start_session/3` means six edits.

**Where:** `test/vutuv_web/live/shell_live_test.exs`, `shell_live_presence_test.exs`, `shell_live_browser_notifications_test.exs`, `browser_tab_teaser_test.exs`, and `test/vutuv_web/organization_message_badge_test.exs`.

**Now:** five of them declare a byte-identical private `session_for/2` — `Sessions.start_session(user, build_conn(), alert: false)` merged into `%{"session_token" => token}` — and the sixth inlines it twice. That shape is the one issue #1036 pinned down, so it is exactly the thing likely to change again. Two further copies sit in `section_reorder_live_test.exs` and `reference_check_live_test.exs`.

**Want:** one `shell_session(user, extra \\ %{})` in `test/support/conn_case.ex`, beside `create_and_login_user/2`. Per-file defaults stay at the call site.

jpawlowski:
**Symptom:** the mount session for the off-router `live_render` children — the `"session_token"` key and the `alert: false` flag — was spelled out in seven test files. A change to `Vutuv.Sessions.start_session/3` meant seven edits.

**Change:** one `shell_session(user, extra \\ %{})` in `test/support/conn_case.ex`, called by all seven — `ShellLive`, `SectionReorderLive` and `ReferenceCheckLive` alike. Per-file defaults stay at the call site. `checkbox_checked?/2` moves out of the `using` quote with it: the quote sat at exactly credo's 150-line `LongQuoteBlocks` ceiling, so nothing could be added. Both are plain functions on `ConnCase` now, imported the way `Vutuv.DataCase` does with `errors_on/1`.

**Not included:** `ignore_comments: true` on that credo check would give the quote ~44 lines back, but it is repo-wide config.

**Verified:** test-only, no behaviour change. `mix precommit` gates green; the ten affected test files 259/259.

Closes #1779.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Vutuv.MixProject do
   def project do
     [
       app: :vutuv,
-      version: "7.560.3",
+      version: "7.569.2",
       elixir: "~> 1.20",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
--- a/test/support/conn_case.ex
+++ b/test/support/conn_case.ex
@@ -154,16 +154,18 @@ defmodule VutuvWeb.ConnCase do
   # it at 150 lines — it sat at exactly 150 before this pair moved out.
 
   @doc """
-  The mount session a browser hands `VutuvWeb.ShellLive`.
+  The mount session a browser hands the off-router `live_render` children —
+  `VutuvWeb.ShellLive`, `SectionReorderLive`, `ReferenceCheckLive`.
 
   A real, active session's raw token under `"session_token"` — the only
-  identity key the shell trusts (issue #1036), so a test drives it the way a
-  request does instead of curating a map (the curated map is the *production*
-  path, `VutuvWeb.LayoutHTML.shell_session/1`). `alert: false` keeps the
-  new-device security notice out of the test's mailbox.
+  identity key those LiveViews trust (issue #1036), so a test drives them the
+  way a request does instead of curating a map (the curated map is the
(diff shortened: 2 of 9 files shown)


wintermeyer:
Keeping this — six copies of the mount session down to one is worth it, and moving the pair out of the `using` quote to get under credo's 150-line ceiling is the kind of thing that only gets harder later.

It is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — silent, not red. Its version line says 7.483.3, which is now *behind* `main`, so rebase and re-bump; a version below main is the one collision `bump_version.exs` cannot warn you about.

Yes to the two you left out (`section_reorder_live_test.exs`, `reference_check_live_test.exs`) — fold them in on the rebase rather than opening a second PR for two call sites.

jpawlowski:
Folded both in — `section_reorder_live_test.exs` and `reference_check_live_test.exs` now call `shell_session/2`. They passed the test's own conn, which is why they sat out; `start_session/3` reads only user-agent and `remote_ip` from it, and both files hand over the unmodified ConnCase conn (`nil` / 127.0.0.1), so it was never a difference. The docstring names all three children again.

On the version: already re-bumped — the branch is at 7.518.1, above main's 7.507.2, and 0 commits behind. It merged (not rebased) four times since your read; a rebase would need a force-push, so I left that alone.

PR #1774, opened 2026-08-27
docs(contributing): a pull request title may say what reviewing costs [2m ?]

jpawlowski:
**Symptom:** Twenty open pull requests, no reading order. The diff size misleads: #1743 listed `+2878`, of which 1820 lines were `priv/gettext/` — 933 lines of code.

**Change:** One bullet in `CONTRIBUTING.md`. Conventional Commits plus an optional bracket: review minutes, `?` when a product decision is needed rather than a code review, `after #N` for a real code or schema dependency. It also names the scope vocabulary, because `ui` hides whether a reviewer needs a phone. The value of a change stays out — that is the reviewer's judgement. Documentation only: no script, no Action, no label bot. Hot deploy.

**Out of scope:** whether the `type(scope):` prefix should also go on our German commit subjects — a second, larger proposal.

**Verified:** our open pull requests already carry the format. #1738 merged with its title unchanged, bracket and all.

Closes #1770.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1792, opened 2026-08-28
perf(build): a docs-only push skips the test suite [2m ?]

### LINKED ISSUE #1791: Skip the test suite on a push that only changes Markdown
**TL;DR** Every push pays the full ~9,100-test suite, including a push that changes nothing `mix precommit` can read.

**Where:** `.claude/hooks/precommit-before-push.sh` ends in an unconditional `mise exec -- mix precommit`.

**Now:** Measured 2026-08-27/28: ~300 s at load 6, ~900 s at load 72 (twenty worktree sessions share this machine). #1774 changes one Markdown file and a version line, and has paid that run nine times — each conflict autofix pushes again. No run said anything about the change.

**Want:** Before the run, list the files the push newly puts on the remote. If all of them are files precommit provably cannot read — `docs/`, `.github/`, top-level Markdown, `LICENSE` — skip it and say why on stderr. Anything else, and any doubt (detached HEAD, no merge base, unreadable list), runs the full gate. It removes work instead of adding machinery; CI still checks everything.

jpawlowski:
**Symptom:** every push runs `mix precommit`, whose cost is its ~9,100-test suite: ~300 s idle, ~900 s when twenty worktree sessions share the machine. #1774 changes one Markdown file and has paid that nine times — each conflict autofix pushes again. No run said anything about the change.

**Change:** `.claude/hooks/precommit-before-push.sh` lists what the push newly puts on the remote and skips the run only when every file is one precommit provably cannot read: `docs/`, `.github/`, `LICENSE`, `NOTICE`, `CODEOWNERS`, top-level Markdown. Deny-first, so `priv/help/*.md`, `.claude/rules/design.md`, `rel/` and `mix.exs` still pay the gate.

**Not included:** no change to CI, which keeps checking everything. The hook is a pre-check only.

**Verified:** `test/vutuv/precommit_hook_test.exs`, 22 cases, calibrated both ways — removing the exemption reds the skip cases, widening it to any `.md` reds the build-input cases.

Closes #1791.


wintermeyer:
I like the shape of this: deny-first, and every way of not knowing (unreadable arguments, detached HEAD, a foreign refspec, no merge base) falls through to the full run rather than guessing. That is the right direction for a check with an exemption.

One real gap. The description says "top-level Markdown", but the code exempts `.md` at **any** depth outside the seven denied roots — `*` matches `/` in a `case` pattern. I ran the predicate on its own:

```
EXEMPT  content/posts/hello.md
EXEMPT  rel/overlays/notes.md
EXEMPT  native/thing.md
```

Nothing is broken today because none of those directories exist, which is exactly what makes it worth closing now rather than after someone adds one. Two lines: put a `*/*) return 1 ;;` immediately before the `*.md)` arm so only genuinely top-level Markdown is exempt, and add `rel/` to the denied roots — release overlays are a build input and it is the one project directory the list is missing.

Also: it is CONFLICTING, so GitHub cannot build the merge ref and **no CI runs at all** — the PR goes silent rather than red. Rebase onto `origin/main` and re-bump. And since it edits a hook of Stefan's, the merge is his call, not mine.

wintermeyer:
The exemption stops firing once the branch integrates main, which is the case this is built on. I drove `--explain` at each shape:

| push | result |
|---|---|
| first push of a docs-only branch | SKIP |
| further docs edit, branch in sync | SKIP |
| re-push after merging main in | **PUSH** |
| re-push after rebasing onto main | **PUSH** |

`merge-base(@{upstream}, HEAD)` falls behind the integration point, so the diff carries all of main's code. Basing it on `merge-base(origin/main, HEAD)` instead makes all four skip, with every fail-closed case unchanged. I ran both.

wintermeyer:
Pushed the trunk-base change onto this branch. Three things beyond what I described above:

- the diff base is now the merge base with the trunk;
- the `@{upstream}` fallback is gone rather than demoted. It is the permissive reading, and rule 1 of the file says an unanswered question is not an exemption;
- the skip message says "this branch", because that is what is now measured.

Three cases pin it, calibrated: revert the change and exactly those go red. `mix precommit` green here, 9,301 tests.

PR #1764, opened 2026-08-27
fix(posts): take the link teaser off the screenshot worker [5m after #1713]

jpawlowski:
**Symptom:** with `SUMMARIZE_LINKS` on, the teaser sentence was written inside `Screenshots.process/2`, on the screenshot worker — one job at a time, five per poll, 30 s per answer. One drain could spend two and a half minutes on prose while someone who had just posted a link waited on a placeholder for their picture.

**Change:** the drain hands the job to `Vutuv.Posts.SummaryQueue` and returns. That process writes one teaser at a time and stands down while an image scan is open (`ImageScans.busy?/0`, fail-open). The work is `Screenshots.write_link_summary/3`; it writes by id and re-announces the card, so the teaser fills in with no reload.

**Not included:** durability. The queue is in memory, capped at 20, dropping the oldest. A lost job costs the card its second line; a page with its own `og:description` is never queued.

**Merge order:** #1713 first, then this rebases onto `main`.

**Verified:** `mix precommit` green, regression test calibrated both ways.

Closes #1742.

*An AI agent wrote this text in my name. I know that is problematic.*

PR #1798, opened 2026-08-29
fix(fediverse): the entity table, the clamp, and a NUL that reached Postgres [5m]

### LINKED ISSUE #1741: Two of each small helper in the remote-content stack
**TL;DR** #1713 wrote a better version of two shared helpers *beside* the originals instead of into them. Both now exist twice, and the weaker copy still serves its old callers.

**Where:** the ellipsis clamp — `Vutuv.SocialFeed.Post.truncate/2` (5 callers) against `Vutuv.LinkSummary`'s private `clamp/1`; the entity decoder — `Vutuv.RemoteHtml`'s private `decode_entities/1` against `Vutuv.OpenGraph`'s.

**Now:** `LinkSummary`'s clamp trims the trailing partial word, `Post.truncate/2` cuts mid-word. `OpenGraph` decodes named and numeric entities in one pass; `RemoteHtml` runs six `String.replace/3` calls that work only because `HtmlSanitizeEx.strip_tags/1` decoded the rest first, and whose ordering rule (`&amp;` last) is a comment somebody has to keep obeying.

**Want:** decide which of each pair is right, move it into the owning module, delete the copy. Both fixes reach callers outside #1713, which is why they were left out of it.

**Repro:** none — neither is visible to a member today.

jpawlowski:
**Symptom, twice.** A remote server writing `&rsquo;` got `Google&rsquo;s new phone` stored verbatim: `Vutuv.RemoteHtml` decoded entities with six `String.replace/3` calls and an ordering rule kept in a comment. Six entries where the web has two thousand.

And a remote `&#0;` reached Postgres. `strip_tags/1` decodes numeric entities itself, so the byte existed before any guard ran — and this text is stored (`Fediverse.remote_text/3` writes it into a delivered post's body), where `22021 character_not_in_repertoire` is a raise on insert. Any federating server could trigger it.

**Change.** One pass over the full HTML5 table (`:mochiweb_charref`, already a transitive dep); a NUL scrub where the byte appears. `Post.truncate/2` now cuts between words for all four callers, with an O(1) `byte_size` fast path (72,255 → 16,334 reductions).

Also: the three `ugc` profile anchors withhold the referrer like `VutuvWeb.Markdown` already does, and an inverted clock bracket in `checks_test`.

Split out of #1713 at @wintermeyer's request. **#1713 — and #1760 and #1764 behind it — are stacked on this; land this first.** #1713 then promotes the two helpers to public when `Vutuv.OpenGraph` becomes their second caller.

Closes #1741.

### DIFF
--- a/lib/vutuv/remote_html.ex
+++ b/lib/vutuv/remote_html.ex
@@ -13,7 +13,7 @@ defmodule Vutuv.RemoteHtml do
     * `<br>` and `</p>` become the line breaks that carried the meaning,
     * every remaining tag is stripped (`HtmlSanitizeEx.strip_tags/1`), so there
       is no allowlist to get wrong and nothing to render `raw`,
-    * the base entities are decoded exactly **once**,
+    * HTML entities are decoded exactly **once** (`decode_entities/1`),
     * the sending server's own **custom-emoji shortcodes** go out with it
       (`strip_shortcodes/1`), and
     * the result is clamped, so one hostile delivery cannot park a novel.
@@ -87,6 +87,7 @@ defmodule Vutuv.RemoteHtml do
     |> String.replace(~r{<br\s*/?>}i, "\n")
     |> String.replace(~r{</p>}i, "\n\n")
     |> HtmlSanitizeEx.strip_tags()
+    |> scrub_nul()
     |> decode_entities()
     |> String.trim()
     |> strip_shortcodes()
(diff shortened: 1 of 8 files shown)


wintermeyer:
Reviewed properly and merging this. Two things I pushed to your branch, and one correction.

**The NUL is real.** I reproduced it end to end on `main` before touching anything:

```
AFTER_FULL_CHAIN_NUL: true
BYTES: <<104, 101, 108, 108, 111, 0, 119, 111, 114, 108, 100>>
PG_SAYS: REFUSED: ERROR 22021 (character_not_in_repertoire)
         invalid byte sequence for encoding "UTF8": 0x00
```

`RemoteHtml.to_text` → `SearchText.normalize_search` → `Note.changeset` → `Repo.update()`, exactly as you describe. Good find.

**There is a second one you did not catch, and this PR did not fix it.** `strip_tags/1` decodes numeric references itself and its parser builds every one it sees — and `:mochiutf8.codepoint_to_bytes/1` has no clause past `0x10FFFF`. So `&#1114112;` does not become a replacement character, it raises a `FunctionClauseError` from *inside* `strip_tags/1`: one line earlier than `scrub_nul/1`, and an exception rather than a failed insert. On `main` and on this branch alike:

```
OOB_DECIMAL  &#1114112;: {:raised, FunctionClauseError}
OOB_HEX      &#x110000;: {:raised, FunctionClauseError}
HUGE         &#99999999;: {:raised, FunctionClauseError}
```

I added `defuse_wide_charrefs/1` ahead of the parser — it escapes the `&` so the reference stays literal text, which is what mochiweb already does with a lone surrogate. Plus a test covering both spellings, the `0x10FFFF` boundary (which must still decode, or the guard is off by one), and that a surrogate and an ordinary reference are unaffected. Calibrated: pull the pipeline step and it goes red with that exact `FunctionClauseError`.

**The correction.** The entity half of your symptom does not reproduce. `strip_tags/1` already resolves the full HTML5 table and re-escapes only the four HTML-special entities, so `Google&rsquo;s new phone` was never stored verbatim — I measured thirteen entities through the real pipeline on `main` and every one came out right:

```
&rsquo; -> a’b   &mdash; -> a—b   &hellip; -> a…b   &eacute; -> aéb   &#x2019; -> a’b
```

The hand-written six covered exactly what `strip_tags/1` leaves escaped, so it was complete, not deficient. The rewrite is still worth having — nobody can tell that by reading it, and the chain carried an ordering rule (`&amp;` last) that lived in a comment and had to keep being obeyed — but it is a clarity change, not a bug fix, and I rewrote the code comment to say so. Worth knowing for the next PR: the claim was in the body *and* in the comment, so it would have shipped as fact.

`mix precommit` green here, 9,127 tests.

PR #1800, opened 2026-08-29
refactor(mastodon-api): share the Web Push crypto instead of owning it [15m]

jpawlowski:
First half of #1729, split out of #1752 at your request — **land this one first**.

**Where:** `Vutuv.WebPush`, `Vutuv.MastodonApi.WebPush`.

The VAPID and `aes128gcm` crypto lived inside the Mastodon adapter, where a subscription hangs off an API access token. An installed web app has no token, so it could reach none of it without dragging the adapter's `MASTODON_API_ENABLED` gate along — a gate that must not silence a member's own app.

- `Vutuv.WebPush` owns the keys and the delivery; `Vutuv.MastodonApi.WebPush` is a thin delegate that keeps its own gate.
- The changeset rules move to `Vutuv.WebPush.Validations`, so a second schema cannot quietly lose the SSRF check.
- `push/2` owns deleting a subscription the push service reports gone — the promise the dispatcher kept by hand.

Behaviour-preserving: no new table, route or config. #1752 is rebased onto this and carries the service worker, the new table and the settings switch.

### DIFF
--- a/lib/vutuv/languages.ex
+++ b/lib/vutuv/languages.ex
@@ -107,6 +107,27 @@ defmodule Vutuv.Languages do
     |> Keyword.get(:locales, ["en"])
   end
 
+  @doc """
+  The language to tell somebody else this member reads, or `"en"` for a member
+  who never chose.
+
+  Deliberately **not** clamped to `site_locales/0`, which is the difference
+  between naming a locale and rendering in one. `Vutuv.Notifications.Emailer`
+  has to render, so it refuses a locale we hold no translations for; a Web Push
+  payload only *names* the language, and the reader on the other end may have
+  strings we do not — a third-party Mastodon client with a French translation
+  should be told "fr" rather than "en" (`Vutuv.MastodonApi.PushDispatcher`),
+  and our own service worker falls back to its own table on a locale it was
+  given no line for (`Vutuv.WebPush.Dispatcher`).
+
+  vutuv is installable by third parties, so the fallback is the same `"en"`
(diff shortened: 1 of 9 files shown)


wintermeyer:
Reviewed and merging. This is the split done right — the whole diff is the lift, and I could check it as one thing.

I verified the parts that would have been expensive to get wrong:

- **The gate stayed where it belongs.** `MastodonApi.WebPush.enabled?/0` is `MastodonApi.enabled?() and WebPush.enabled?()`, and the shared module carries no Mastodon gate at all — only comments naming it. So `MASTODON_API_ENABLED=false` still silences the client API and cannot silence a member's own installed app, which was the point.
- **Both halves of the SSRF pair survived the move.** `Ssrf.internal_host?/1` in the changeset (cheap, no DNS, safe to run there) and `Ssrf.resolves_to_internal?/1` at send time — the diff moves the send-time check rather than dropping it, which is the one I went looking for. Matching `Vutuv.Webhooks.Subscription` is the right precedent.
- `push/2` returning `:ok` rather than an error when the gate is off is a good call and worth the comment you gave it: since `push/2` now owns deleting a subscription the service reports gone, an error there would have quietly deleted rows for the duration of a config switch.

Pulling `user_locale/1` up into `Vutuv.Languages` with the "naming a locale is not rendering in one" distinction is the kind of thing that would otherwise have been rediscovered as a bug.

I rebased it onto current `main` myself — #1798 and #1799 landed under it and turned it CONFLICTING — resolved only the version line and re-bumped to 7.534.1. `mix precommit` green, 9,128 tests.

PR #1807, opened 2026-08-29
fix(push): send legacy aesgcm so relay-backed iOS clients can decrypt the payload [5m]

### LINKED ISSUE #1698: Stop iOS clients failing to decrypt every push we send

jpawlowski:
**Symptom:** every push from us reaches Ivory on iOS as "Unable to decrypt notification". We always sent `Content-Encoding: aes128gcm`, which keeps the salt and the sender key inside the record. An iOS client sits behind an APNs relay, and APNs forwards a payload rather than our headers, so the relay has to lift those two values out of `Encryption:` and `Crypto-Key: dh=`. With `aes128gcm` there is nothing to lift.

**Change:** new column `mastodon_push_subscriptions.standard`, default false as on Mastodon. `Vutuv.WebPush.content_encoding/1` turns it into the encoding, `encode_body/4` builds it, and the new `encrypt_aesgcm/5` is the second ceremony. The installed app's own subscription carries no flag and gets `aes128gcm`, so #1752 is unaffected.

**Not included:** flipping the default. That needs someone who can confirm the relays no longer need the headers.

**Verified:** both encodings against their specs' own test vectors, plus an assertion that the shipped headers actually open the shipped body. Not verified on a physical device.

Supersedes #1771, which GitHub closed when this branch was renamed; the review is there.

Closes #1698.

### DIFF
--- a/docs/architecture/mastodon-api.md
+++ b/docs/architecture/mastodon-api.md
@@ -713,6 +713,29 @@ subscription and revoking an app takes it along. A push carries **no content**:
 the kind and the notification id, nothing else, so neither the push service nor
 a lock screen learns what was written.
 
+**Two content encodings, and the subscription picks.** `subscription[standard]`
+is stored per device and defaults to false, exactly as on Mastodon, in which
+case the body is the older `aesgcm` with its salt in `Encryption:` and the
+sender key in `Crypto-Key: dh=`. That is not legacy politeness: an iOS client
+subscribes through an APNs relay, and APNs forwards a payload rather than our
+headers, so the relay has to lift those two values out and put them in the
+payload itself — `aes128gcm` hides both inside the record and the phone shows
+"Unable to decrypt notification" (issue #1698). A client that sets the flag gets
+RFC 8291 `aes128gcm`. Both are checked against their specs' own test vectors in
+`web_push_test.exs`.
+
+**The false default is a compatibility decision, not an oversight, and it has to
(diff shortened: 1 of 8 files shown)


jpawlowski:
Closing this: keeping the original branch name and continuing in #1771 instead. The rename was reverted.

PR #1976, opened 2026-09-03
The profile checklist asks three things about the profile

wintermeyer:
The owner's onboarding card on `/:slug` now asks three things, all of them about the profile it sits on.

Out: **"Add a tag"**, which sign-up already requires — it arrived ticked for everybody who came in the front door, a strikethrough line teaching the reader that the list is about things already behind them. And **"Follow other members"**, whose link led out of the card and whose subject is the feed, not the profile; the welcome flow asks it once with real accounts to pick from, and the "Who to follow" rail keeps asking.

In: **the LinkedIn import**, which was a link under the list and is a step in it now. It counts as done off what an import leaves behind (a work experience), not off the act — somebody who typed the job in by hand has done the same thing, and a step only a particular button can tick reads as an advert for that button. The second link under the list, the connections importer, answers "who do I know here"; that is the "Who to follow" card's question and it lives on there.

One test asked for "Add a tag" as text and so also matched "Add a tagline" and the Tags card's own tile. It asks by destination now.

Surface: `/:slug`, owner only, first hour after sign-up.

https://claude.ai/code/session_01KE2NaGdpqUG5idSC7Cf2v1

### DIFF
--- a/lib/vutuv_web/live/user_profile_live.ex
+++ b/lib/vutuv_web/live/user_profile_live.ex
@@ -1330,17 +1330,12 @@ defmodule VutuvWeb.UserProfileLive do
     )
   end
 
-  # Re-derive the checklist from the current assigns; called from the initial
-  # load and from the social-graph refresh (the follow step's count changes
-  # live). Reads :followee_count, so it must run after put_social_assigns.
+  # Re-derive the checklist from the current assigns. Every step now reads the
+  # member's own record, so this no longer depends on the social graph; it stays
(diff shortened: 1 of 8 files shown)


wintermeyer:
Closing this: #2001 shipped the same decision on 5 September, two days after this branch. The tag step and the follow step are gone, the LinkedIn import is a step in the list rather than a link under it, and the contacts importer moved to the "Who to follow" card. `main` goes a little further — the tick counts an education too, the steps jump to their own field, and the list is only built for the owner.

PR #2019, opened 2026-09-06
Tell a member what was claimed, who decided, and how long they have

### LINKED ISSUE #2010: The mail about a hidden post does not say what was claimed or how long you have
When a report hides a member's post, the mail says a contribution was reported and links the case page, which shows the category and the text; nothing says what the reporter claimed, that the hiding was automatic, whether the ground is the law or the house rules, or how long they have. The notice, the mail and the case page should carry all of it: the category, the reporter's explanation quoted (never who wrote it), one line that the content was hidden automatically when the report arrived, the ground, and the three options with the 72-hour deadline. Every member whose content is reported notices; this is the statement of reasons the Digital Services Act asks for (Art. 17). It waits on #2008 for the category and the required explanation; nothing waits on it.

Where: `Vutuv.Moderation.Notifier.owner_content_frozen/1` and `moderation_case/show.html.heex`

wintermeyer:
A member whose post is hidden by a report was told the category and shown their own text, and nothing else: not what the reporter claimed, not that no person decided it, not the ground, not how long they have. Worse since #2018, the mail still promised everyone that an edit brings the content straight back, which is wrong for a copyright case.

All three surfaces now carry the same statement of reasons from one source, `Vutuv.Moderation.owner_notice/1`: the category, the reporter's explanation quoted and never their name, the automatic freeze, the ground (house rules or copyright law), and the options that case really has. `owner_edit_offer/2` answers the edit question once instead of seven templates recombining two booleans, so a message or job posting is not offered an editor it has no button for, and the review mail promises no 72-hour round it does not have.

Entry points: `/moderation/cases/:id`, `Vutuv.Moderation.Notifier`, `moderation_frozen_*` / `moderation_review_*` (three locales, both formats). Also corrects a wrong function name in `docs/architecture/moderation.md`.

Closes #2010

Shipped. The notice, the mail and the case page now all say what was claimed, quote the reporter's words without naming them, state that the freeze happened automatically the moment the report arrived, name the ground as the house rules or copyright law, and list only the options that case actually has, with the 72-hour deadline where there is one. The options are answered once by `Moderation.owner_edit_offer/2` rather than recombined per template, because the copyright case from #2018 had already made the old "an edit makes it visible again" promise false and no locale should be able to get that wrong on its own. I left the wording of the moderation *digest* untouched beyond delegating its one moderation line to the same sentence, and did not widen the case page's edit button to job postings, which is a separate decision.

### DIFF
--- a/docs/architecture/email.md
+++ b/docs/architecture/email.md
@@ -83,6 +83,28 @@ blank line inside the quote gets the bare `">"` — a quote block ends at the
 first line without a marker, and mail clients that strip trailing whitespace
 would turn `"> "` into exactly that.
 
+**Both halves ask `UserHelpers.split_lines/1` where a line ends, and it is
+written as the effect rather than as a list of spellings.** `Regex.split(~r/\R/u,
+text)` is PCRE's closed set of mandatory line breaks, which is the set UAX #14
+gives a mail client. Enumerating `\r\n`, `\r` and `\n` — the obvious spelling,
+and what shipped first — leaves VT, FF, NEL (U+0085), LINE SEPARATOR (U+2028)
+and PARAGRAPH SEPARATOR (U+2029) inside one "line": invisible in a form, a real
+break in Gmail, Apple Mail and Thunderbird, so everything after one arrives
+**without** the `"> "` that is the whole safeguard. Nothing strips control
+characters from a report note on the way in, so this split is the only thing
+standing between a member in good standing and a forged vutuv signature with a
+sign-in link inside a DKIM-signed vutuv mail — on a site where signing in means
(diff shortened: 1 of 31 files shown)


wintermeyer:
Not merging yet — one thing first. CI is green and the rest checked out, but `email_quoted_text/1` splits on `\r\n`, `\r`, `\n` only. A note containing U+2028, U+2029, U+0085, VT or FF therefore keeps the `> ` off every line after it, and Gmail, Apple Mail and Thunderbird all break at those. That is exactly the spoof the docstring says the prefix prevents. `Regex.split(~r/\R/u, text)` covers the set; the new test wants a U+2028 payload too.

PR #2072, opened 2026-09-08
The outbox stops counting a post a takedown hid, and three loose ends beside it

### LINKED ISSUE #2069: Three loose ends the copyright takedown check turned up
Three separate small things the end-to-end check of the copyright takedown turned up, each in its own place.

**A hidden post is still counted for other servers.** The machine-readable list of somebody's public posts answers "one" for an author whose only post has been taken offline, while the post itself is gone. Only the number leaks, and the neighbouring list of pinned posts already excludes hidden ones, so the two disagree.

**The declaration a notice needs is thrown away.** A copyright notice is only accepted if the notifier declares in good faith that the use is unlicensed, and that declaration is validated and then discarded. Nothing on the case, in the queue or in any export records that it was made, which is the part of the notice that makes it one.

**The admin's clock is two hours out.** The case page prints the report time and the owner's deadline as bare machine timestamps in universal time, while the history directly beneath shows the same moment in local time. An admin reading the deadline off the header is two hours wrong, and the format is not the one this project uses for a date a person reads.

Where: `Vutuv.Fediverse.public_post_count/1`, `Vutuv.Moderation.Report`, and the admin case page's header

### LINKED ISSUE #2071: The rule against a self-contradicting decision letter is never asked
The rule that a decision letter may never say two contradicting things about the same content exists, and nothing asks it. It is called only from a test, and that test hands it the answer it expects rather than the one the system produced, so it compares a sentence with itself. A tenth way of closing a case was written to try it: the letter went out with a subject saying the content was deleted above a line saying it is merely hidden, and the rule would have said no had anybody asked. It belongs one line inside the function that sends the letter, where both halves are already side by side.

Nothing anybody reads today is wrong; all nine ways of closing a case agree. This is about the tenth, and about a document that currently promises a guard the code does not have.

Where: `Vutuv.Moderation.consistent_outcome?/2`, against `Notifier.deliver_outcomes/3`

wintermeyer:
An author whose only post a takedown had frozen still answered `"totalItems": 1`
to every server that asked, while the note itself 404ed. Three smaller things the
same check turned up ride along.

- **Outbox count** (`Vutuv.Fediverse.public_post_count/1`, and its page twin)
  goes through `Posts.scope_visible/2` now, the SQL twin of the gate deciding
  whether the note is served. That also covers an unreachable author.
- **The good-faith declaration** was validated and dropped. It is a column now
  (additive migration) and a chip on the admin case page.
- **That page's header** printed bare UTC stamps, two hours off the history
  beneath it. Every instant on it reads `<.local_time>` now.
- **`Moderation.consistent_outcome?/2`** was asked only by a test that fed it the
  expected answer. `Notifier.deliver_or_refuse/4` asks it now: a contradicting
  letter is refused and logged, not sent.

Smoke-tested in German in a browser, and against the real outbox.

Closes #2069
Closes #2071

Three separate corrections, all shipped. The outbox count now routes through the
same visibility scope the note itself is served by, which also closes the
unreachable-author case you did not name; the good-faith declaration became a
nullable timestamp column and a chip beside the report's category; and the case
header's three stamps read in the admin's own clock, as does the severance line
one card below that I found on the way.

The guard is asked in `Notifier.deliver_or_refuse/4` now, where the subject and
the fate paragraph meet. When it fires nothing is sent and the claim on
`outcome_notified_at` is handed back, so no row falsely reads as answered, and
the log names the case and both halves. Review caught the guard being stricter
than the contradiction it names: "revised" rules out only `:removed`, so an
owner hidden by an unrelated case editing their post had a true letter refused.
Nothing re-drives an un-claimed row yet, which is #2073.

### DIFF
--- a/docs/architecture/fediverse.md
+++ b/docs/architecture/fediverse.md
@@ -912,6 +912,17 @@ None of this is a guarantee. Remote deletion is advisory by protocol and always
 will be, so the UI never claims a copy is gone — but every takedown now at least
 asks.
 
+**A frozen post is out of the count as well as out of the collections**
+(issue #2069). `Vutuv.Fediverse.public_post_count/1` and its page twin
+`organization_public_post_count/1` are what the count-only `outbox` reports, and
+they filtered a member's audience denials and nothing else — so an author whose
+only post a takedown had hidden still answered `"totalItems": 1` to every server
+that asked, while the note itself 404ed. It is the one number about a hidden
+post that still left the building: the neighbouring `featured_posts/1` already
+serves the anonymous public view, and NodeInfo's `localPosts` counts through
+`Posts.scope_visible/2`, so the outbox was the one surface that disagreed with
+the rest. Both counts now exclude `posts.frozen_at`.
+
 ## Leaving: 410 Gone, and saying so first
 
 Switching the opt-in off used to make every actor endpoint answer `404`, which
--- a/docs/architecture/moderation.md
+++ b/docs/architecture/moderation.md
@@ -39,13 +39,24 @@ things follow from that.
 
 **It is only accepted complete.** `Report.changeset/3` requires the note (which
 work, where the original can be seen) and a good-faith declaration whenever the
-category is `copyright`. The declaration is a **virtual** field, not a column:
-the changeset refuses the category without it, so a stored copyright report *is*
-the record that it was made, and a column would be a second copy free to drift.
-The check sits in the changeset rather than in the controller, so the Mastodon
-report API cannot file half a notice either (it has no field for the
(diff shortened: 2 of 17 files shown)


wintermeyer:
Checked this and did not merge. CI is green, but the guard also refuses a truthful letter: `consistent_outcome?("revised", fate)` demands `fate == :visible`, so an owner whose account is hidden (frozen by another case, or unreachable) editing reported content closes `resolved_edited` over a measured `:hidden` — and that reporter is then never told, permanently. Reproduced on both. The exhaustive table in `moderation_notice_wording_test.exs` already models `revised` as contradicting only `:removed`; the code is stricter than its own test.

Issue #2066, opened 2026-09-08
Taking a picture offline in a second checkout hides it from the first one too


wintermeyer:
Taking a picture offline in a second checkout moves the file out of the shared store and into that checkout, so every other checkout loses the picture and the sweeper that puts held files back cannot find it any more. Tearing that checkout down then takes the file with it. The hold directory a takedown uses should be shared the way every other picture directory is, which is a single line in the list the setup script walks.

The neighbouring hold, the one the automatic picture check uses while it decides, is already shared. Two directories doing the same job disagree, which reads as an omission rather than a decision. Anybody working in more than one checkout notices, and today that is every agent session; on a single-checkout installation nothing changes.

Where: `CANONICAL_TREES` in `scripts/restore-snapshot.sh`, against `Vutuv.Uploads.hold_dir/1`

*An AI agent wrote this text in my name. I know that is problematic.*

PR #2076, opened 2026-09-08
Stop telling agents to delete the upload symlinks

wintermeyer:
The smoke-test rule told anybody working in a second checkout to symlink the upload directories and then delete the links again before running the gate. Both halves were wrong.

The links are infrastructure, created by `scripts/restore-snapshot.sh` and re-created nightly. `:uploads_dir_prefix` is empty in dev, so every image path resolves against the running server's own directory; a checkout without them serves 404s for every picture. Six agents followed the deletion advice during one milestone before another session pointed at the script's own header.

The test the rule named as the early warning does not warn either. It lstats the link rather than reaching through it, so it ran green twice with all thirteen in place; "beyond a symbolic link" never fires when the target is inside the same repository.

What is dangerous is narrower and is now what the rule says: an upload as an existing member replaces that member's served sizes and their private original, and no derivation brings the original back. That destroyed a real avatar on 2026-09-07, and a snapshot does not carry `originals/`, so restoring one cannot undo it. Drive picture smoke tests as a throwaway account. `frozen/` is the one unlinked tree, recorded with #2066.


wintermeyer:
Merged as 0084979db — the merge landed while GitHub answered 502, so the pull request never updated its own state. Closing by hand.

Issue #2083, opened 2026-09-09
A press photo or logo is a picture kind of its own on the shared image table


wintermeyer:
Every uploaded picture is re-encoded for the screen and its original stays private, so nothing on vutuv can hand out a print-quality photo or a vector logo. A `press_kit` kind should be born on the shared `images` table, never a table of its own, owned by a member or a page, carrying a caption (Markdown like a work-experience description, so the photographer's `@handle` links to their profile and notifies nobody), credit, position, the rights confirmation's timestamp and a logo flag, its files at `press_kit/<token>/…` proxy-served like post photos, so no nginx change, and its caps (ten photos, 30 MB, the logo variants) as config with an ADMINS.md entry. Nobody outside the code notices; every other sub-issue waits on it. A photo is JPEG, PNG or WebP, exactly the containers the metadata stripper can take apart, so the download is always the cleaned original and never leaks GPS or a serial number; a logo is SVG or PNG, the SVG handed out only as an attachment with `nosniff` beside a PNG rendering. Two traps: the row is born pending and invisible until the scan releases it, and the proxy root is a new reserved slug, like `post_images`.

Where: `Vutuv.Images`, `Vutuv.Uploads.Spec`, `Vutuv.Uploads.MetadataStrip`

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Built in PR #2091. Notes for the sub-issues that build on this.

**One claim in the body is wrong.** "The proxy root is a new reserved slug, like `post_images`" contradicts the project rule that no feature may claim a new root path word (profiles own `/:slug`). The proxy is at **`/system/press_kit/:token/:version`**, the pattern `RemoteMediaController` set, and `Vutuv.Accounts.ReservedSlugs` is untouched. Two other body claims checked out: production serves proxied images with `send_file`, not the X-Accel handoff (`:post_image_serving` in `runtime.exs`), so "no nginx change" is true; and `MetadataStrip.supported?/1` really is exactly `.jpg .jpeg .png .webp`, which is now the photo whitelist by construction rather than by coincidence — `.heic` is refused even where libvips can decode it.

[...]

PR #2113, opened 2026-09-10
An organization page offers the same press section, managed by its team

### LINKED ISSUE #2087: An organization page offers the same press section, managed by its team
A page has one logo, re-encoded for the screen, and nothing a journalist could download. The page should get the same card and section page, the editor reachable from the page's manage menu beside roles and domains for owners and publishers, and the page's existing logo offered as the first logo variant with one click when it was uploaded as SVG. Page teams and whoever writes about the organization notice; it waits on the profile sub-issue. The rows are the page's, not the uploader's: `organization_id` owns them, the uploader sits in `uploader_user_id` and is nulled when their account goes, the way the logo survives today, so a press photo never vanishes with the colleague who uploaded it. The trap is the visibility check: a pending picture must be visible to every owner and publisher of the page, not only to the member who uploaded it, or a colleague cannot see what is being checked.

Where: `VutuvWeb.OrganizationLive.Show`, `Vutuv.Organizations.image_visible_to?/2`

wintermeyer:
A page had one screen-sized logo and nothing a journalist could download, while
`Vutuv.PressKit` has taken either owner kind since #2083.

The card, the section page, the four agent-format siblings, the sitemap entry
and the editor are now the same code with an organization as the owner. The
editor is `/organizations/:slug/press/edit` — the member's own
`VutuvWeb.PressKitLive`, embedded rather than routed — on the manage tab bar for
an **owner or publisher**. A page whose logo is a vector takes it onto the logo
shelf with one press, and every team member sees a picture the AI check still
holds, not only whoever uploaded it.

Underneath: `PressKit.page_path/1` built a page's URL from its root handle,
which serves the bare `/:slug` only. It answered a 404.

On top of #2087, on your word: the credit field falls back to the account's
name — the page's, for a page's kit.

Closes #2087

https://claude.ai/code/session_01WPMGFtmCZNip8EJiRrx6ch

### DIFF
--- a/docs/architecture/images.md
+++ b/docs/architecture/images.md
@@ -936,6 +936,32 @@ siblings are `VutuvWeb.AgentDocs.PressKitDoc`, and the profile document carries
 the same entries under `press_kit`; both list the **released** pictures only,
 since a document that names a file must name one that can be fetched.
 
+**A page has the same section, kept by its team (issue #2087).** The card, the
+section page, the documents, the schema.org block and the sitemap entry are the
+same code with an organization as the owner — `PressKitComponents.press_card/1`
+(lifted out of the profile template when the page needed the second copy),
+`VutuvWeb.PressKitController.organization/2`, `PressKitDoc.build/2`. Three things
+are the page's own. **The URL is built from the slug**, `/organizations/:slug/press`
+(`PressKit.page_path/1`), not from a claimed root handle: that handle dispatches
+the bare `/:slug` alone, so `Identity.path/1 <> "/press"` answered an address that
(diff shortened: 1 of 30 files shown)


wintermeyer:
Not merging yet. Two handlers were not revisited when `:owner` became a second principal. I measured both with a publisher whose role is revoked while the editor stays open.

`delete` (press_kit_live.ex:226) never re-asks, so the picture and its original were purged, irreversibly. `save` reaches `first_error(:forbidden)` and kills the socket — unreachable on `main`, where owner was always the viewer. `move`, `reorder` and upload refuse correctly.

Everything else in the review held.

Issue #2129, opened 2026-09-10
The card offers tags that are suddenly busy on other servers


wintermeyer:
When a conference runs or something happens, a topic goes from nothing to hundreds of posts within the hour, and a member here hears of it only if they already follow that tag. Every Mastodon server publishes its trending tags with a seven-day history, without a login and for one request each, so a row under "In your feed right now" can offer what is spiking elsewhere, one click to follow it. On the day this was written `#warntag` stood at 1,045 uses against a median of 3 on the days before, trending on seven of the eight servers asked.

The trap is that the loudest tag is often a machine. `#mow4` trended on all eight the same day: 40 of 40 posts from one domain, all 40 flagged as bots, all within ninety seconds. Both defences come free with the answer we already fetch, since a server marks bot accounts itself and the spread of authors across domains sits in the same response. A spike also shortens that tag's fetch interval at once, rather than waiting for the measured rate to catch up.

Where: the trending endpoint of each configured server, beside the fetcher from the sibling issue.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
<img width="362" height="493" alt="Image" src="https://github.com/user-attachments/assets/49c14e1b-e011-4bdf-8b6e-7bec07ac5c40" />

wintermeyer:
Built in PR #2159 (branch `claude/external-tag-trending-2129`).

**The hypotheses in the body.** The headline one holds and is now measured: on 10 September 2026 `#warntag` stood at **5,994 uses across the ten shipped servers against a median of 25** on the six days before, and it trended on **10 of the 10** that answered at `limit=20` (the body's "1,045 uses, seven of eight" was one server's figure and one day's list). `#mow4` is real too, and worse than described: it trended on **9** servers, and its timeline on troet.cafe was **40 of 40 statuses from `social.prepedia.org`, 40 of 40 flagged as bots** — not "40 of 40 within ninety seconds", the forty spanned 8.6 hours. Two claims do **not** hold:

- **"Both defences come free with the answer we already fetch" is false.** `/api/v1/trends/tags` carries `name`, `url` and seven days of `{day, accounts, uses}` — no author, no domain, no bot flag. Both defences need the tag's own timeline, which is a second request. So the pass spends one `GET` per server plus one census per candidate it is about to offer (capped at eight).
[...]

Issue #2127, opened 2026-09-10
What the other servers found stands in the feed like any other post


wintermeyer:
Posts nobody reads are worth nothing in a cache. A tag follow already pulls that tag's posts into the member's feed, so the finds from other servers belong in the same place: an ordinary card carrying the author's own address, a quiet line naming the server it was found through, and a link to the original. Hidden words and muted tags apply as everywhere else, a blocked instance never appears, and every card can be reported. The same rows fill the fediverse tab of a tag page, which for most topics stands empty today.

One thing to get right: the server we asked is not the author's home. A post about Koblenz found through troet.cafe was usually written somewhere else entirely, so the card names the author's server and mentions ours only in the small line. Anyone reading a post has to be able to tell whose words these are and where they live.

Where: the feed assembly in `Vutuv.Posts`, `Vutuv.Tags.Timeline`, the post card in `VutuvWeb.PostLive.Feed`.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Built in PR #2151 (branch `claude/external-tag-cards-2127`), for whoever picks up #2128–#2129.

**The hypotheses in the body.** Two of the three hold, one does not. `Vutuv.Tags.Timeline` is where the body says and gained a third union arm. The post card is in `VutuvWeb.PostComponents`, not in `VutuvWeb.PostLive.Feed` — the feed only calls it — and the new `<.external_post_card>` sits beside `remote_post_card/1` and reuses its skin (`remote_avatar`, `remote_header`, `remote_body`, `card_menu`, `hide_list`). And "the feed assembly in `Vutuv.Posts`" is **not** a query: the member feed is nine independent fetchers merged in Elixir by `Vutuv.FeedPage`, so this is an eighth source in `feed_sources/3` rather than a fourth arm of anything.

[...]

Issue #2125, opened 2026-09-10
A followed tag remembers which servers it should come from


wintermeyer:
A tag follow is a member and a tag, with nowhere to say where the posts should come from. It should carry sources, one row each: `vutuv` plus any server the member picked. A table rather than a list on the follow, because the fetcher's question is the other way round — which server-and-tag pairs does anybody here want? — and that is one grouped query instead of unpacking every member's array. Nobody sees this issue directly; the card and the fetcher stand on it.

The migration writes exactly one source, `vutuv`, to every follow that exists, so nothing changes for anybody until they add a server themselves. It is a pure addition and needs no second deploy. Organization pages follow tags too and get the same treatment.

Where: `Vutuv.Tags.TagFollow`, `Vutuv.Tags.follow_tag/2`.

*An AI agent wrote this text in my name. I know that is problematic.*

wintermeyer:
Built in PR #2134 (branch `claude/external-tag-sources-2125`), for whoever picks up #2126–#2129.

**The hypotheses in the body.** `Vutuv.Tags.TagFollow` and `Vutuv.Tags.follow_tag/2` are both where the body says. The opening sentence is not: since #1336 a follow row is a member **or** an organization page (nullable pair, CHECK exactly one), and pages come in through a second entry point, `follow_tag_as_organization/2`. Both now write the local source, in one shared transaction.

**What shipped.** `tag_follow_sources`, one row per source. The column is `source`, not `host`: the local row is no hostname, and a caller building `https://#{source}/tags/…` out of it would fetch nonsense. A pasted address of our own becomes `vutuv` rather than a server to poll (`Vutuv.Fediverse.own_host?/1`, so `www.` and the tag host count). The server-name grammar is no longer copied — `BlockedInstance` publishes `host_format/0` and `max_host/0`. Context: `add_tag_follow_source/2`, `remove_tag_follow_source/2`, `tag_follow_sources/1`, `wanted_tag_sources/0`, `local_tag_follow_source/0`.

[...]

PR #2138, opened 2026-09-10
A followed tag brings posts from the servers it names

### LINKED ISSUE #2126: Posts carrying a followed tag arrive from the servers it names
Nothing fetches anything yet. A worker should read the tag timeline of each server a follow names, keep the newest handful as plain text plus a link to the original, and skip what is sensitive or behind a content warning. The rows belong in a small table of their own with a cap per tag and a hard ceiling, not in `fediverse_posts`: a cached remote post drags image ingestion, the AI gate, the screenshot queue and two counting sweepers behind it, and a REST status is not an ActivityPub object to begin with.

How often is the tag's own business: fetch so that roughly five new posts accumulate, at least ten minutes apart and at most three hours. Measured, that puts a tag in the middle of a news event at ten minutes and a quiet local one at three hours, and both drift back on their own. A budget per server keeps twenty busy tags from spending its rate limit, and the clock is stamped on every outcome, skips included, or a pair that can never be fetched holds the front of the queue forever.

Where: a sweeper beside `Vutuv.Fediverse.CountsRefresher`, HTTP through `Vutuv.SocialFeed.Http`.

wintermeyer:
Following a tag showed what members here wrote about it, and nothing from the servers a follow named since #2125. A background job now reads each named server's public tag timeline (`GET /api/v1/timelines/tag/:hashtag`) and keeps the newest posts as plain text plus a link — never a picture, so nothing reaches the AI image gate — skipping anything sensitive or behind a content warning.

The rows go in `external_tag_posts` / `external_tag_fetches`, not `fediverse_posts`. A pair's clock is stamped on every outcome, skips included, so an unfetchable pair cannot hold the front of every batch (#1316). Three refusals sit at fetch time, where the changeset cannot reach: the instance blocklist, `Ssrf.vetted_address/1` with the connection pinned, and a name that yields no hashtag.

`FETCH_EXTERNAL_TAG_POSTS=false` asks nobody anything. No user-visible surface yet — the cards are #2127–#2129.

Where: `Vutuv.Tags.ExternalPosts`, `Vutuv.Tags.ExternalTagClient`.

Closes #2126

https://claude.ai/code/session_01AW11tx7rYiSEaRDjPFNghY


wintermeyer:
Blocking on review: `remote_id`, `author_acct` and `author_name` validate at 255 **graphemes** against `varchar(255)`, which counts codepoints. 100 ZWJ emoji pass the changeset at 700 codepoints and raise 22001 inside `store/2`, outside `fetch/2`'s rescue. Measured: the raise aborts the whole pass, no pair is stamped, and every later run raises again. That is #1316, installation-wide. `Handle.display_name/1` does not strip emoji, so an ordinary display name reaches it.

PR #2153, opened 2026-09-10
A member decides once in settings whether machines may read their posts

### LINKED ISSUE #2107: A member decides once in settings whether machines may read their posts
Nothing keeps search engines and AI crawlers away from a member's posts, and asking per post would be one question too many in the composer. A member should set it once in their settings; every new post keeps the answer it was published with, so changing the setting later leaves older posts exactly as they were. A post that says no is also not delivered to the Fediverse at all, because a remote server keeps its copy forever and no header reaches it — the settings page has to say that plainly, since the member is deciding it once rather than post by post. Every member notices, and so does anybody reading their posts through a machine.

The other half is unchanged: the composer asks whether to remove metadata from attached files (author name, software, timestamps out of a PDF, via `qpdf`), default on, hidden when qpdf is absent or older than 11.10.0.

Where: `VutuvWeb.ContentPolicy`, the settings hub, `Vutuv.Posts`.

wintermeyer:
Nothing kept search engines and AI away from a member's posts, and a post that said no leaked its whole body wherever it sat inside somebody else's — most sharply on `/:slug`, which is public, anonymous, in the sitemap and carries no `X-Robots-Tag`.

`/settings/privacy` now holds one switch, *Search engines and AI may read my posts* (`Vutuv.Prefs`' `posts_machines_allowed?`, default on). `Vutuv.Posts` stamps the answer onto `posts.noindex_noai?` as a post is published, so flipping it later leaves older posts exactly as they went out.

Three review rounds each measured completely and each missed the next surface, so this is enforced as a property rather than a list: a post's body leaves the database toward a machine-readable surface only through `Posts.scope_machines_for/2`, and `test/vutuv/post_body_chokepoint_test.exs` fails the build when a new query selects a `body` past it. Exceptions carry their reason. A withheld post also leaves the Fediverse, the sitemap and every listing document entirely; the HTML conversation still shows it to people.

**To decide:** page and doc differ on purpose, and `/ModDate` survives the strip.

Closes #2107

https://claude.ai/code/session_013VnEsuePUe2CB2QvDcshcp

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,11 @@ jobs:
       # pdftoppm renders the qualification proof-document PDFs
       # (Vutuv.QualificationDocument); without it the PDF test paths
       # capability-skip and lose their coverage.
-      - name: Install poppler-utils (PDF rendering) and ffmpeg (video on posts)
-        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends poppler-utils ffmpeg
+      # qpdf joins them for issue #2107: it is what takes the author's name,
+      # the software and the dates out of a PDF, and several fixtures already
+      # built their hostile PDFs with it and skipped themselves without it.
+      - name: Install poppler-utils (PDF rendering), ffmpeg (video on posts) and qpdf (PDF metadata)
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends poppler-utils ffmpeg qpdf
 
       - name: Set up Elixir/OTP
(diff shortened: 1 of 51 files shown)


wintermeyer:
I'm closing this one rather than rebasing it. It was parked twice because the chokepoint guards queries while the leak comes from preloaded ancestors, so it still hands crawlers a withheld body. It also bundles unrelated attachment-metadata work. #2107 and #2183 stay open, and the branch stays for reference.

PR #2176, opened 2026-09-11
Reporting a foreign post takes every copy its own server vouches for

### LINKED ISSUE #2164: Reporting a foreign post leaves its copies standing, and the message says otherwise
Reporting a foreign post asks "Unsere Kopie wird sofort für alle auf diesem vutuv gelöscht?" and then does neither half of that. Measured: the reported row is blanked and stamped rather than deleted, and the identical copies of the same post that were fetched from the other servers keep standing on the page — one status was six rows, one of which was marked. The member sees the post they just reported still there, under a different "found through" line, and has been told it is gone for everyone.

Either the sentence should describe what happens, or reporting should take every copy of that original. The second is the honest reading of what a member means when they press it. Anybody who reports a foreign post notices immediately, which is the worst moment to look untrustworthy.

Where: the `report-external-post` handler in `VutuvWeb.PostLive.Feed` and `VutuvWeb.TagLive.Timeline`, and `Vutuv.Tags.ExternalPosts`.

wintermeyer:
Reporting a post one of your followed tags pulled from another server blanked
the clicked row and left the identical copies from the other servers standing,
under a sentence promising they were gone.

This re-lands #2169, reverted as `8b2c1a862`. That version related the copies by
the post's address plus its author host, and the polled server writes both while
any member may add any host as a tag source. A planted card could blank honest
copies of somebody else's post, and a tombstone planted first kept it out of the
table for good.

The guard: only a row whose polled server, permalink host and author host all
name one server may reach copies it did not file; every other row reaches the
rows its own server filed. The ingest gate asks the same question, so a forged
tombstone refuses nothing. Two sentences now say which happened.

Where: `Vutuv.Tags.ExternalPosts.reaches?/2`, a card's ⋯ menu on `/feed`.

Closes #2164

https://claude.ai/code/session_01AW11tx7rYiSEaRDjPFNghY

### DIFF
--- a/docs/architecture/social-graph.md
+++ b/docs/architecture/social-graph.md
@@ -284,13 +284,27 @@ it again on the way in, so a tombstone can only refuse what its own report could
 have blanked and a planted one cannot keep an honest post out of the table.
 `report/2` answers `{:ok, :every_copy}` or `{:ok, :this_copy}` because the
 member has to be told which of the two happened; the confirm dialog and the
-flash are two sentences for that reason. Measured over 78 rows of a production
-copy, the author's half of the key costs nothing (43 groups either way) and the
-authority half narrows 51 of the 78 possible reports to the clicked row — the
-relayed copies of posts whose author's server nobody here asked, and every one
-of them a row a stranger could equally have invented. The first attempt at this
-keyed the takedown on the description alone, shipped as `7cdfd4dc7` and was
-reverted as `8b2c1a862` within the hour.
+flash are two sentences for that reason. What the authority half costs is that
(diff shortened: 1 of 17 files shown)


wintermeyer:
Red three times in a row, each time in the attachment tests — `message_attachment_web_test.exs:196`, then `:160`, then `chat_attachments_test.exs:198` and `:152`. This branch touches no attachment code and neither test file. It is #2186: the preview page of a text attachment is drawn by headless Chromium under a 30 s deadline, and a loaded runner misses it.

Waiting for #2186 rather than re-running again, then rebasing.

PR #2217, opened 2026-09-14
The member directory finds a member by where they worked or studied

oliverandrich:
Knowing where somebody worked or studied was no help in finding them. The member directory searched names and nothing else, while the two CV sections that carry exactly that sit public on every profile, in every agent document and in the sitemap — everywhere except the one box a visitor types into.

/system/members now has two more checkboxes, Firma and Schule & Uni, over work_experiences.organization (a linked public page's name too) and educations.school. Every entry counts, so a role left in 2016 answers like the current one, and such a row names the matching entry rather than today's employer, which would say nothing about why it is in the list. Nothing becomes newly visible; what changes is that it is findable.

Each word's match set is a UNION of one-table queries rather than an OR, and that is the part not to tidy up later: inside one OR spanning tables Postgres uses none of the trigram indexes and scans users, and all five boxes start ticked, so every visitor would pay that per keystroke. The measurements, and the case the shape does not fix, are on Directory.field_match/2. The migration adds three GIN indexes CONCURRENTLY, additive and N-1 safe.

To decide: a word under three characters still scans, and the date-range formatter now has one home instead of three.


wintermeyer:
Would you mind adding a screenshot?

oliverandrich:
Sure.

Screenshot-2026-09-15-jDJQKenU@2x

wintermeyer:
Thank you, Oliver! Your union query and the three trigram indexes now power the site search: #2267 finds members by employer and school on `/search`, with organizations as a kind of their own, and credits you as co-author. I'm closing this one because we moved the feature from five checkboxes on the member directory to the search page, which already has the scopes and operators people know.

PR #2243, opened 2026-09-18
fix: preserve picker order for post photos

### LINKED ISSUE #2158: The cover of a multi-photo post is the picture you picked last
Picking several photos for a post puts the cover on the wrong one. The composer appends each picture the moment its upload finishes rather than in the order they were picked, so the smallest file arrives first and wears the Cover badge: four photos picked largest-first came back smallest-first, with the cover on a 23 kB thumbnail the author chose last. Whoever posts more than one picture at a time notices, which on a phone's camera roll is most of them.

It is milder than it sounds, because the author sees the order before posting and can drag it. But the default is wrong in the one direction that matters, since the cover is what the feed and every share preview show. The order the file picker hands over is the order the author meant.

The Media Kit editor had the identical defect and is being fixed; the fix there writes each upload's rank down on first sight of the entry list rather than on completion. The same shape applies here, with one difference worth checking: the composer uploads automatically, so an entry can finish before the list is next read.

Where: `VutuvWeb.PostLive.Composer`, around `composer.ex:1743`

mvanhorn:
Start with regression tests through the real composer upload path, then record unseen entry refs in picker order at the start of every image progress callback, before checking `done?` or consuming any entry; the existing `PressKitLive.note_slots/2` demonstrates the necessary timing. Keep a small composer-local ordered list of tagged pending refs and completed image IDs, seeded from current images, append newly observed refs after existing reservations, and replace a successful ref with its image ID so `assigns.images` can be projected in reserved order without waiting for all uploads. Multi-photo posts default to upload-completion order rather than file-picker order, making a small photo selected last become the cover. In `Composer.handle_progress(:images, entry, socket)`, successful uploads currently append to `assigns.images`, and the save path passes that list directly as `image_ids`.

Register three valid JPEG uploads in picker order, keep them in flight, finish them in reverse, and assert the rendered hidden `post[image_ids][]` order and cover badge belong to the first selected image. Submit and use the existing `only_post/1` helper to assert reloaded image IDs and positions preserve selection order; Follow the established multi-client pattern in `test/vutuv_web/live/press_kit_live_test.exs`: separate `file_input` clients registered before completion, with partial `render_upload` calls, avoid LiveViewTest closing all channels of a multi-entry client when the first entry is consumed. Also cover multiple registered entries where the first progress callback completes a later entry immediately, without an extra validate event.

Fixes #2158


wintermeyer:
@mvanhorn Do you have a vutuv account? I wasn't able to find you.

wintermeyer:
Thanks a lot, Matt! The smallest photo becoming the cover had been bugging me, and your tests pin it down well: three uploads in flight, finishing in reverse. Against the old composer nine of them fail with exactly that reversed order. I merged main into your branch so CI ran against today's code, and it's deploying now.

mvanhorn:
Thanks for merging, @wintermeyer, and for the kind words on the tests. Recording the entry refs before the `done?` check was the whole fix.

